PR #3, opened 2017-01-05
Added command line option to exclude window title from events

ErikBjare:


johan-bjareholt:
How does this work with the webapp visualization?

Does it totally break or does it simply just not show any titles?

ErikBjare:
@johan-bjareholt I have no clue, I'll investigate before merging.

johan-bjareholt:
Or you could just set the title as "Unknown", "Hidden" or something similar

ErikBjare:
@johan-bjareholt I did as you said and set the title as "title:excluded".

ErikBjare:
Merging this, should be problem-free.

PR #6, opened 2017-01-31
Switched to the Heartbeat API

ErikBjare:
### Depends on the following PRs
 - https://github.com/ActivityWatch/aw-client/pull/6
 - https://github.com/ActivityWatch/aw-watcher-window/pull/5

### DIFF
--- a/aw_watcher_window/main.py
+++ b/aw_watcher_window/main.py
@@ -9,7 +9,7 @@ from aw_core.models import Event
 from aw_core.log import setup_logging
 from aw_client import ActivityWatchClient
 
-from . import lib
+from .lib import get_current_window
 
 logger = logging.getLogger("aw.watchers.window")
 
--- a/aw_watcher_window/main.py
+++ b/aw_watcher_window/main.py
@@ -15,9 +15,6 @@ logger = logging.getLogger("aw.watchers.window")
 
 
 def main():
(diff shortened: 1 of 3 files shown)


aw_watcher_window/main.py
```diff
@@ -82,59 +39,23 @@ def main():
...
+        now = datetime.now(timezone.utc)
+        if current_window is None:
+            logger.debug('Unable to fetch window, trying again on next poll')
+        else:
+            # Create current_window event
+            labels = ["title:" + current_window["title"]]
+            labels.append("appname:" + current_window["appname"])
+            current_window_event = Event(label=labels, timestamp=now)
+
+            client.heartbeat(bucketname, current_window_event, 2*args.poll_time)
```
johan-bjareholt: The default poll time is 1.0, what does that mean? A second? A minute?

And why 2*poll_time?

ErikBjare: One second. The `2*poll_time` is there since the loop block obviously takes more than `poll_time`seconds (we know it takes at least that long due to the `sleep(poll_time)`). 

A better value would probably be `poll_time + 1`, I'll fix that tomorrow.

johan-bjareholt: isn't 1 second a little too low as the default value?

PR #7, opened 2017-03-10
Added verbose flag, added check for DISPLAY environment variable on startup

ErikBjare:


ErikBjare:
Straight forward enough. Merging.

PR #9, opened 2017-03-14
Fixed crash when starting quickly

johan-bjareholt:
xprop.py assumed that xprop always would find _NET_ACTIVE_WINDOW, which
can be false. Usually when no window is focused it becomes 0x0, but that
variable seems to be defined upon window unfocus so if aw-watcher-window
starts before any window has been focused it used to crash.

### DIFF
--- a/aw_watcher_window/xprop.py
+++ b/aw_watcher_window/xprop.py
@@ -34,16 +34,16 @@ def xprop_root() -> str:
 def get_active_window_id():
     lines = xprop_root().split("\n")
     match="_NET_ACTIVE_WINDOW"
-    result = False
+    result = None
     for line in lines:
         if match in line:
             result = line
             break
+    wid = "0x0"
     if result:
         wids = re.findall("0x[0-9a-f]*", result)
-    wid = "0x0"
-    if len(wids) > 0:
-        wid = wids[0]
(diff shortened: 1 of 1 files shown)


aw_watcher_window/xprop.py
```diff
@@ -33,8 +33,14 @@ def xprop_root() -> str:
...
+    result = None
+    for line in lines:
+        if match in line:
+            result = line
+            break
+    if result:
+        wids = re.findall("0x[0-9a-f]*", result)
     wid = "0x0"
     if len(wids) > 0:
         wid = wids[0]
```
ErikBjare: len(wids) could fail since wids could be undefined (if `result = None`). Try setting `result = False` before the `if result:` line. (I haven't tried this myself)

johan-bjareholt: >>> if None:
    ...   print("you are wrong")
    ... else:
    ...   print("you are right")
    ... 
    you are right

johan-bjareholt: Ugh, i understood it wrong and then wrote a false test case...

I'll fix this and go to sleep lol

ErikBjare:
Looking good, go ahead and merge when Travis is done.

PR #11, opened 2017-04-21
Added Makefile

ErikBjare:


Makefile
```diff
@@ -1,2 +1,9 @@
+.PHONY: build
+
 build:
-	- gcc -framework Carbon get_window.c
+#	- gcc -framework Carbon get_window.c
```
johan-bjareholt: What's this?

ErikBjare: Something old we did when trying to get it running on macOS, not needed anymore.

PR #12, opened 2017-04-22
Cleanup and switch back to python-xlib

ErikBjare:
Relying on having xprop installed is kind of nasty, so I cleaned up the xlib version and switched to it (ended up a lot cleaner than xprop). 

Seems to work fine, but not extensively tested.


ErikBjare:
@johan-bjareholt Could you please test and review this?

ErikBjare:
Yay, lets do this.

PR #20, opened 2017-07-18
Make watcher shutdown as aw-qt parent process dies.

nikanar:
https://github.com/ActivityWatch/aw-qt/issues/19

This makes the child process (the watcher) check it's PPID, so that if ever its parent process dies (aw-qt), this one shutdowns also. 1 is `PPID` of init process, what PPIDs get set to when parents die (aka "the orphanage").

Sorry for the commit fluff, git still gets the better of me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: build
 
 build:
-	pip3 install . --process-dependency-links --user
+	pip3 install . --process-dependency-links
 
 test:
 	python3 -c "import aw_watcher_window"
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: build
 
 build:
-	pip3 install . --process-dependency-links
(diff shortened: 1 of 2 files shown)


johan-bjareholt:
Great work! 👍 

Are committing with another user than on GitHub or did someone else do the commits? Personally I don't care since it looks great anyway, but next time if you want to top the [contributors list](http://activitywatch.net/contributors/) I would recommend setting your email in your git config to the same as your github account so the commits are linked to you.

> Sorry for the commit fluff, git still gets the better of me.

No problem, it's a small PR so i can just squash and merge so they look like one complete commit later :)

ErikBjare:
Just a note: This has the same issue with PyInstaller as I mentioned in https://github.com/ActivityWatch/aw-watcher-afk/pull/24

nikanar:
It's me, holding on wrong config files. Thanks :)

Issue #21, opened 2017-07-20
UnicodeDecodeError in xlib.py 


nikanar:
Some internet page title caught the aw-window-watcher by surprise :

```
2017-07-20 16:20:19,793 [INFO ]: No longer AFK  (aw_watcher_afk.afk:99)
2017-07-20 16:30:45,707 [INFO ]: Became AFK  (aw_watcher_afk.afk:105)
2017-07-20 16:31:05,738 [INFO ]: No longer AFK  (aw_watcher_afk.afk:99)
2017-07-20 17:51:01,673 [ERROR]: Exception thrown while trying to get active window: 'bytes' object has no attribute 'encode'  (aw_watcher_window.main:73)
Traceback (most recent call last):
  File "/home/will/.local/lib/python3.6/site-packages/aw_watcher_window/xlib.py", line 62, in get_window_name
    return d.value.decode('utf8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 40: invalid start byte

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/will/.local/lib/python3.6/site-packages/aw_watcher_window/main.py", line 70, in heartbeat_loop
    current_window = get_current_window()
  File "/home/will/.local/lib/python3.6/site-packages/aw_watcher_window/lib.py", line 44, in get_current_window
    return get_current_window_linux()
  File "/home/will/.local/lib/python3.6/site-packages/aw_watcher_window/lib.py", line 14, in get_current_window_linux
    name = xlib.get_window_name(window)
  File "/home/will/.local/lib/python3.6/site-packages/aw_watcher_window/xlib.py", line 64, in get_window_name
    return d.value.encode('utf8').decode('utf8')
AttributeError: 'bytes' object has no attribute 'encode'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/aw-watcher-window", line 11, in <module>
    load_entry_point('aw-watcher-window==0.2', 'console_scripts', 'aw-watcher-window')()
  File "/home/will/.local/lib/python3.6/site-packages/aw_watcher_window/main.py", line 45, in main
    heartbeat_loop(client, bucket_id, poll_time=args.poll_time, exclude_title=args.exclude_title)
  File "/home/will/.local/lib/python3.6/site-packages/aw_watcher_window/main.py", line 74, in heartbeat_loop
    traceback.print_exc(e)
[...]

PR #23, opened 2017-07-28
check out for bytes instead of string when window title fails to decode

nikanar:
https://github.com/ActivityWatch/aw-watcher-window/issues/21

> we need a check to avoid the .encode('utf8') if d.value is a bytes object.

I'm not sure if `d.value` can *not* be a bytes object at that point, after a failed `decode()` (so whether the `else` branch ever gets visited), I only trust the writer of the former version who tried to `d.value.encode()`, from which I deduce that at it is possible sometimes (therefore, sometimes `d.value` is not bytes).


ErikBjare:
Looks good to me, merging. Thanks @nikanar!

Issue #19, opened 2017-06-29
Got BadWindow exception while trying to get active window on Linux


ErikBjare:
Got this while running aw-watcher-window (commit 7612c4a9377ad9daa274e0242b09d59fba830d7c):

    2017-06-28 12:36:49,972 [INFO ]: Running watcher with poll time 1.0 seconds  (aw_watcher_window.main:35)            
    2017-06-28 12:36:49,973 [INFO ]: aw-watcher-window has started  (aw_watcher_window.main:43)                         
    2017-06-28 12:36:50,019 [INFO ]: Connection to aw-server established  (aw_client.client:251)                        
    2017-06-28 12:36:50,030 [INFO ]: Loaded 1 failed requests from queuefile  (aw_client.client:236)                    
    2017-06-28 20:43:26,199 [WARNING]: Unable to get window class, got a BadWindow exception.  (root:74)                

    2017-06-28 20:43:26,199 [WARNING]: Code made an unclear branch  (root:79)                                           
    2017-06-28 20:43:26,202 [ERROR]: Exception thrown while trying to get active window: <class 'Xlib.error.BadWindow'>: code = 3, resource_id = <class 'Xlib.xobject.resource.Resource'>(0x00c00004), sequence_number = 43768, major_opcode = 15, minor_opcode = 0  (aw_watcher_window.main:64)
    Traceback (most recent call last):                        
      File "/home/erb/Programming/activitywatch/aw-watcher-window/aw_watcher_window/main.py", line 61, in heartbeat_loop
        current_window = get_current_window()                 
      File "/home/erb/Programming/activitywatch/aw-watcher-window/aw_watcher_window/lib.py", line 44, in get_current_window
        return get_current_window_linux()                     
      File "/home/erb/Programming/activitywatch/aw-watcher-window/aw_watcher_window/lib.py", line 13, in get_current_window_linux
        cls = xlib.get_window_class(window)                   
      File "/home/erb/Programming/activitywatch/aw-watcher-window/aw_watcher_window/xlib.py", line 80, in get_window_class
        window = window.query_tree().parent                   
      File "/home/erb/Programming/activitywatch/venv_main2/lib/python3.6/site-packages/python_xlib-0.19-py3.6.egg/Xlib/xobject/drawable.py", line 425, in query_tree
[...]

PR #24, opened 2017-08-01
Catching Xlib errors

nikanar:
Fixing https://github.com/ActivityWatch/aw-watcher-window/issues/19

Check for Xlib's `XError`, covering both BadValue and BadWindow (among others). If so, raise a warning instead of an error, and let the heartbeat return an "unknown" window.

And in case of any other error, print it instead of failing.

- - - 
Asides re automated checks :

Re the failing `make test`, I used the same `import Xlib` as in [xlib.py](https://github.com/ActivityWatch/aw-watcher-window/blob/master/aw_watcher_window/xlib.py) so I don't quite understand why it complains about it.

The code quality watcher seems genuinely mistaken about formatting, as it recommends `%`-formatting, which [everyone](https://stackoverflow.com/questions/12382719/python-way-of-printing-with-format-or-percent-form) says is soon-to-be deprecated, especially for python 3  (instead of `"{}".format()`).

### DIFF
--- a/aw_watcher_window/main.py
+++ b/aw_watcher_window/main.py
@@ -67,7 +67,7 @@ def heartbeat_loop(client, bucket_id, poll_time, exclude_title=False):
             logger.debug(current_window)
         except Exception as e:
             logger.error("Exception thrown while trying to get active window: {}".format(e))
-            traceback.print_exc(e)
+            traceback.print_exc()
             continue
 
         now = datetime.now(timezone.utc)
--- a/aw_watcher_window/main.py
+++ b/aw_watcher_window/main.py
@@ -6,6 +6,8 @@ import os
 from time import sleep
 from datetime import datetime, timezone
 
+import Xlib
+
 from aw_core.util import assert_version
 from aw_core.models import Event
 from aw_core.log import setup_logging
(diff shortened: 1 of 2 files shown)


aw_watcher_window/main.py
```diff
@@ -65,9 +67,12 @@ def heartbeat_loop(client, bucket_id, poll_time, exclude_title=False):
         try:
             current_window = get_current_window()
             logger.debug(current_window)
+        except Xlib.error.XError as e:
```
johan-bjareholt: Please don't catch Xlib.error errors here. The main.py file is used by mac and windows aswell which do not have the xlib dependency, so please try to catch the exceptions somewhere in the xlib.py file instead.

nikanar: Ah I get it now, thanks. Will do.

nikanar:
As noted in a comment, I'm not sure what to do in `get_window_name`, return some kind of "unknown" value already, or keep trying with `window.get_wm_name()` (so return None at the error point). Depends on what the root cause is, and whether `window.get_wm_name()` is likely to succeed or not, but since we have only two occurrences of this event per week, it's tedious to test myself.

With the proposed change, if as I expect `window.get_wm_name()` fails next, then I suggest to replace `d = None` by `return "unknown"`. If it succeeds and we never hear of XErrors in get_window_name again, good for us.

ErikBjare:
@nikanar Don't pay too much attention to Landscape, we haven't configured it properly.

It complained about `import Xlib` on the macOS build, where Xlib isn't installed (for obvious reasons). Catching the exception in instead xlib.py fixed it.

ErikBjare:
Are we ready to merge? :)

PR #29, opened 2018-02-11
Update MacOS title logic

### LINKED ISSUE #28: No titles on MacOS
Not sure if this is a known issue, but the window watcher always reports empty titles for MacOS (Sierra). I'd be happy to help debug, I'd like to get involved with ActivityWatch development.

ianobermiller:
- Switch from invoking osascript to pure Python
- Clean up the api to the macos.py file (have it return the dict directly)
- Fix a typo in README

Fixes #28 

Test Plan:
- source .././venv/bin/activate
- aw-watcher-window --testing --verbose
- verify the app name and title being printed are correct in a few different applications


README.md
```diff
@@ -1,12 +1,10 @@
...
 Cross-platform window-Watcher for Linux (X11), macOS, Windows.
 
 [![Build Status](https://travis-ci.org/ActivityWatch/aw-watcher-window.svg?branch=master)](https://travis-ci.org/ActivityWatch/aw-watcher-window)
 
 ### Note to macOS users
 
-To log current window title the teminal needs access to macOS accessibility API.
+To log current window title the terminal needs access to macOS accessibility API.
 This can be enabled in System Preferences > Security & Privacy > Accessibility, then add the Terminal to this list. If this is not enabled the watcher can only log current application, and not window title.
-
```
ianobermiller: Oops, the extra changes here were due to prettier, I believe. Let me know if you want them reverted. Not sure why it would remove the trailing line...

ianobermiller:
Ok, that took entirely too long. The issue was that I had moved the app around, and ended up with the accessibility permissions disabled.

ErikBjare:
Hey, this looks great. Does it work well? I'd rather do it like this than with the [`printAppTitle.scpt`](https://github.com/ActivityWatch/aw-watcher-window/blob/master/aw_watcher_window/printAppTitle.scpt).

ianobermiller:
One downside is that it requires MacOS Yosemite or higher. But I do agree
the JavaScript is much nicer. If we are ok with that limitation I’ll clean
this up.

ErikBjare:
The Yosemite limitation is fine. 

However, I'm curious if running it as JS would have a significant performance impact since you'll be spinning up a JS VM on every call. Would be interesting to see how it compares to the `.scpt`.

exoji2e:
One way to reduce the load might be to keep the VM alive? I think we need to benchmark this. But yes the `.scpt`-solution we did not choose out of beauty, but rather because this was the only solution I found after some days of searching the web.

johan-bjareholt:
I am trying to understand how osascript works but then started wondering, wouldn't it be possible to find the underlying OS API call which these applescript/javascript files by using dtruss? (the mac alternative to strace)
It's not a request for this PR, but maybe in the future if we have some spare time :)

ianobermiller:
Did some digging, it wasn't too bad to do with Python only :D

aw_watcher_window/macos.py
```diff
@@ -1,17 +1,28 @@
...
 
+def getInfo() -> Optional[dict]:
+    app_name = ''
+    title = ''
 
-def getInfo() -> str:
-    cmd = ["osascript", os.path.join(os.path.dirname(os.path.realpath(__file__)), "printAppTitle.scpt")]
-    p = subprocess.run(cmd, stdout=PIPE)
-    return str(p.stdout, "utf8").strip()
+    app = NSWorkspace.sharedWorkspace().activeApplication()
```
exoji2e: `activeApplication` is deprecated: https://developer.apple.com/documentation/appkit/nsworkspace/1530411-activeapplication, 

and is replaced by `frontmostApplication`: https://developer.apple.com/documentation/appkit/nsworkspace/1532097-frontmostapplication

aw_watcher_window/macos.py
```diff
@@ -1,17 +1,28 @@
...
+                # We could use app['NSApplicationName'], but this value is more
+                # accurate and matches other methods (like applescript)
+                app_name = window['kCGWindowOwnerName']
+                title = window.get('kCGWindowName', u'')
+                break
 
-
-def getTitle(info) -> str:
-    return info.split('","')[1][:-1]
+    return {"appname": app_name, "title": title}
```
exoji2e: Not sure if it ever happens that the pid is nto found in `window_list`, but maybe it can occur because of realtime problems. We should probably move the return to the break, so we don't get a `name-error` here.

ErikBjare:
The Travis build is failing because the `aw-watcher-window.spec` file used by PyInstaller references the now removed `.scpt`. 

Just remove the relevant tuple from the `datas` argument of `Analysis` and it should work.

exoji2e:
Some more thoughts:

I think `AppKit` inside python is included if you install Xcode. We might want to check for it in the makefile, and [install pyobjc](https://stackoverflow.com/questions/12767669/import-error-no-module-named-appkit) otherwise.

However I'm guessing the most problems will occur with the binary release for macOS-users, since those might not have Xcode, not really sure what to do about that.

ErikBjare:
@exoji2e It should be possible to include the right pyobjc dependency in the PyInstaller build.

I remember we used something like this before, can't remember why we switched but probably worth investigating.

@ianobermiller Don't feel discouraged by the above. It might take a while to merge your commits, but we will do it sooner or later. If you don't have the time to look into it: just let us know and we'll build on your commits when the time comes.

ianobermiller:
I have zero experience with Python installs, so it would probably be a poor
use of my time :). Let me know if there is more I can do.

ErikBjare:
Closing, work continued in #40

PR #37, opened 2019-10-06
Make exclude_title a config option

cunidev:
Works fine here, but as my knowledge of the codebase is limited any in-depth testing is welcome.

### DIFF
--- a/aw_watcher_window/config.py
+++ b/aw_watcher_window/config.py
@@ -6,10 +6,12 @@ from aw_core.config import load_config as _load_config
 def load_config():
     default_client_config = ConfigParser()
     default_client_config["aw-watcher-window"] = {
-        "poll_time": "1.0",
+        "exclude_title": True,
+        "poll_time": "1.0"
     }
     default_client_config["aw-watcher-window-testing"] = {
-        "poll_time": "1.0",
+        "exclude_title": True,
+        "poll_time": "1.0"
     }
 
     # TODO: Handle so aw-watcher-window testing gets loaded instead of testing is on
--- a/aw_watcher_window/config.py
+++ b/aw_watcher_window/config.py
@@ -6,11 +6,11 @@ from aw_core.config import load_config as _load_config
 def load_config():
     default_client_config = ConfigParser()
     default_client_config["aw-watcher-window"] = {
-        "exclude_title": True,
+        "exclude_title": False,
         "poll_time": "1.0"
     }
     default_client_config["aw-watcher-window-testing"] = {
-        "exclude_title": True,
+        "exclude_title": False,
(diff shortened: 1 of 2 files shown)


aw_watcher_window/config.py
```diff
@@ -6,10 +6,12 @@
 def load_config():
     default_client_config = ConfigParser()
     default_client_config["aw-watcher-window"] = {
-        "poll_time": "1.0",
+        "exclude_title": True,
```
johan-bjareholt: This changes the default from false to true, the default should be false

aw_watcher_window/config.py
```diff
@@ -6,10 +6,12 @@
...
+        "poll_time": "1.0"
     }
     default_client_config["aw-watcher-window-testing"] = {
-        "poll_time": "1.0",
+        "exclude_title": True,
```
johan-bjareholt: This changes the default from false to true, the default should be false

aw_watcher_window/main.py
```diff
@@ -42,7 +42,7 @@ def main():
     logger.info("aw-watcher-window started")
     sleep(1)  # wait for server to start
     with client:
-        heartbeat_loop(client, bucket_id, poll_time=args.poll_time, exclude_title=args.exclude_title)
+        heartbeat_loop(client, bucket_id, poll_time=args.poll_time, exclude_title=args.exclude_title or config.getboolean("exclude_title"))
```
johan-bjareholt: This is incorrect, an cli argument should always take precedence over a config setting.
You should handle it in the same way as the poll-time argument, pass the config value to parse_args and only use it as the default.

cunidev:
Sorry, I've corrected that. Feel free to test the new commits

aw_watcher_window/main.py
```diff
@@ -42,20 +42,21 @@ def main():
...
     parser.add_argument("--testing", dest="testing", action="store_true")
-    parser.add_argument("--exclude-title", dest="exclude_title", action="store_true")
+    parser.add_argument("--exclude-title", dest="exclude_title", action="store_true", default=default_exclude_title)
     parser.add_argument("--verbose", dest="verbose", action="store_true")
     parser.add_argument("--poll-time", dest="poll_time", type=float, default=default_poll_time)
     return parser.parse_args()
 
 
 def heartbeat_loop(client, bucket_id, poll_time, exclude_title=False):
+    print(exclude_title)
```
ErikBjare: Looks like you missed this one :)

cunidev:
Oh, someone discovered my top secret debug tools :P

johan-bjareholt:
;Looks good!

PR #41, opened 2020-03-28
WIP: Ensure Accessibility permissions on macOS

xylix:
Screenshot 2020-03-28 at 17 43 21
Use AppKit to figure if we have accessibility perms and pyobjc to ask for them.

Still needs to:
- [ ] ask with an actual buttoned dialog, for example something with an accept button that links to the correct settings page, a decline button and maybe a learn more link 
- [x] If user declines not spam them with prompts, and instead just ask once per app lifetime


pyproject.toml
```diff
@@ -14,6 +14,8 @@ aw-client = {git = "https://github.com/ActivityWatch/aw-client.git"}
 python-xlib = {version = "^0.26", platform = "linux"}
 pypiwin32 = {version = "223", platform = "win32"}
 wmi = {version = "^1.4.9", platform = "win32"}
+pyobjc-framework-ApplicationServices = "^6.2"
+pyobjc-framework-CoreText = "^6.2"
```
ErikBjare: These need to have `platform` set, otherwise it will fail to install on other platforms.

xylix:
> Is it possible to add a button on the alert that opens the accessibility settings? I think I've seen that somewhere.

I found these https://macosxautomation.com/system-prefs-links.html which work as links, but I'm still looking for a way to make the button press "open" the link to the accessibility settings. A standard <a href> link didn't work, but there should be a sane way to link in NSAlert buttons.

pyproject.toml
xylix: Good point.

aw_watcher_window/macos.py
```diff
@@ -1,20 +1,35 @@
...
 import os
+import time
 
 
 def getInfo() -> str:
-    accessibility_permissions = AXIsProcessTrusted()
-    #accessibility_permissions = False
+    #accessibility_permissions = AXIsProcessTrusted()
+    accessibility_permissions = False
     if not accessibility_permissions:
```
ErikBjare: I feel like this shouldn't be called in `getInfo()`, since if you miss the message it'll block all tracking until you close it. 

Instead, I'd run it in a separate thread upon start, before the main loop starts.

xylix: Yeah fair point. I'll go for that implementation.

xylix:
Using a native python thread causes a ```objc.error: NSInternalInconsistencyException - NSWindow drag regions should only be invalidated on the Main Thread!

Assertion failed: (NSViewIsCurrentlyBuildingLayerTreeForDisplay() != currentlyBuildingLayerTree), function NSViewSetCurrentlyBuildingLayerTreeForDisplay, file /BuildRoot/Library/Caches/com.apple.xbs/Sources/AppKit/AppKit-1894.30.142/AppKit.subproj/NSView.m, line 13568.
fish: 'python aw_watcher_window/__main…' terminated by signal SIGILL (Illegal instruction)```Error. Will need to look into using macOS's own work queue (https://developer.apple.com/documentation/dispatch/1452927-dispatch_get_global_queue?language=objc )

ErikBjare:
If the issue is that the stuff needs to happen on the main thread, try
using the Python multiprocessing module? A bit hacky but will probably
work.

xylix:
Seems like it should work. I'll look into it later this week.

aw_watcher_window/macos.py
xylix: resolved in f032ed1ad2a7903f42f3077d74a6d2c4f9b440f4

xylix:
Multiprocessing got rid of the `objc.error: NSInternalInconsistencyException - NSWindow drag regions should only be invalidated on the Main Thread!` problem. The alert is still a bit wonky (it doesn't "pop up" in the desktop like an alert should, instead it stays in the background and shakes in the macOS dock.

It does now block execution though... which is weird. The window watcher continues execution after I close the modal.

I think I resolved the conflict correctly, will need to see what CI says.

aw_watcher_window/macos.py
```diff
@@ -9,9 +9,38 @@ def getInfo() -> str:
...
-def getTitle(info) -> str:
+def getTitle(info: str) -> str:
     return info.split('","')[1][:-1]
+
+
+def background_ensure_permissions() -> None:
+    from multiprocessing import Process
+    permission_process = Process(target=ensure_permissions, args=(()))
+    permission_process.start()
+    permission_process.join()
```
ErikBjare, line 24: This is where you block execution!

aw_watcher_window/main.py
```diff
@@ -27,6 +27,10 @@ def main():
     if sys.platform.startswith("linux") and ("DISPLAY" not in os.environ or not os.environ["DISPLAY"]):
         raise Exception("DISPLAY environment variable not set")
 
+    if sys.platform == "darwin":
+        from . import macos
```
ErikBjare, line 31: Maybe do this after logging has been set up? Just in case.

xylix: Fair point. I moved it there to have it closer to the other system independent code in main.py but putting it after logging is a better idea

aw_watcher_window/macos.py
xylix: Oh right of course thats what .join() is meant to do. Whoops. Fixed in b829e04

aw_watcher_window/main.py
xylix: Fixed in b829e04

xylix:
There might have still been some problem, I'll need to manually test on macOS before I can be sure. I'll get that done tomorrow, bit busy today.

xylix:
Looks fine to merge, I tested it locally and solved the conflict of the lock file.

ErikBjare:
@xylix Are the two things in the top comment resolved?

xylix:
Oh right. I guess the other one is not, because it is actually pretty hard in recent macOS to set a setting from dialog. Instead the dialog just gives a message and `Open Accessibility Options` button.

ErikBjare:
@xylix Should we just leave it as is then and merge?

xylix:
Yeah I think we should. The other open PR requires more work before it is ready but this one will be an improvement even in it's current form.

ErikBjare:
Nice! Merging.

Issue #18, opened 2017-06-22
Wayland Support - A blessing and a curse


ErikBjare:
Wayland will complicate matters, we will now have to support whatever each compositor uses.

[Ubuntu will use Wayland from 17.10](http://www.omgubuntu.co.uk/2017/04/ubuntu-will-run-wayland-default) so that would potentially put a lot of users outside of our abilities to collect windows.

Gnome support might be doable, we'll have to check out how you get the active window there. I don't use Gnome and probably wont fix support anytime soon, but it needs done sooner or later.

SanketDG:
Hi,

I use gnome with wayland, and exept things like google-chrome, vlc nothing else seems to get logged. They are simply marked with "unknown"

ErikBjare:
Yeah, sorry about that. You'll have to switch to Gnome with X11 if you want to log the active window.

Unfortunately, I can't do much about it. There is simply no way in Wayland to grab the current window (due to valid security concerns), so we can't watch it.

SanketDG:
Okay, I could switch back to X11. But most distros are switching to Wayland, right? How are you planning to go around this?

ErikBjare:
Yes, most distros using Gnome are switching to Wayland it would seem. We're trying to find a solution, but we can't do it without the help of desktop environment developers. I've, among other things, made a note of it in the Gnome bugtracker.

It will probably take a while, but you can keep an eye on:

 - This issue: https://github.com/ActivityWatch/activitywatch/issues/92
 - A [detailed question on Stack Overflow](https://stackoverflow.com/questions/45465016/how-do-i-get-the-active-window-on-gnome-wayland) where I ask for an answer to the problem.

SanketDG:
I see how the WSM could be immensely helpful to this, hopefully it happens way sooner!

In the meantime, what do you think about adding a note about this somewhere in the docs?

ErikBjare:
I'll add a note of it in the FAQ, it might take a while to update: https://activitywatch.readthedocs.io/en/latest/faq.html

Ramblurr:
Found activity watch, and was interested in using it. But I'm Fedora where wayland is the default. 

So I'll be watching this issue for updates. Good luck :\ seems like this might be a very difficult nut to crack.

ErikBjare:
[...]

PR #46, opened 2020-06-10
RFC: add gnome-shell compatibility

FlorianLudwig:
Add gnome-shell specific way of getting current window. Allows the watcher to work on wayland.

This is gnome-shell specific and won't work on any other wayland based compositor and therefor does not solve #18 

The poetry.lock needs updating but I have to figure out how to do it without updating all libs and just add the new dependency.

As this features requires a new dependency and is specific to one OS with one desktop environment, I marked that dependency as optional. Especially since it requires system libs to be installed to function, increasing the overall footprint quite a bit on non GNOME systems.

### DIFF
--- a/aw_watcher_window/lib.py
+++ b/aw_watcher_window/lib.py
@@ -16,6 +16,7 @@ class Linux:
             import gi.repository.GLib
             try:
                 self.gnome_shell = self.bus.get("org.gnome.Shell")
+                self._setup_gnome()
             except gi.repository.GLib.Error:
                 self.gnome_shell = None
 
@@ -24,20 +25,27 @@ class Linux:
             return self.get_current_window_gnome_shell()
         
         return self.get_current_window_x11()
+    
+    def _setup_gnome(self) -> None:
+        js_code = """
+        global._aw_current_window = () => {
+            var window_list = global.get_window_actors();
+            var active_window_actor = window_list.find(window => window.meta_window.has_focus());
+            var active_window = active_window_actor.get_meta_window()
+            var vm_class = active_window.get_wm_class();
+            var title = active_window.get_title()
+            var result = {"title": title, "appname": vm_class};
+            return result    
+        }
+        """
+        ok, result = self.gnome_shell.Eval(js_code)
+        if not ok:
+            raise Error("failed seting up gnome-shell function: " + result)
 
     def get_current_window_gnome_shell(self) -> dict:
         """get current app from GNOME Shell via dbus"""
-        js_code = """
-        var window_list = global.get_window_actors();
-        var active_window_actor = window_list.find(window => window.meta_window.has_focus());
-        var active_window = active_window_actor.get_meta_window()
-        var vm_class = active_window.get_wm_class();
-        var title = active_window.get_title()
-        var result = {"title": title, "appname": vm_class};
-        result
-        """
 
-        ok, result = self.gnome_shell.Eval(js_code)
+        ok, result = self.gnome_shell.Eval("global._aw_current_window()")
         if ok:
             result_data = json.loads(result)
             return result_data
--- a/aw_watcher_window/lib.py
+++ b/aw_watcher_window/lib.py
@@ -1,19 +1,65 @@
 import sys
+import json
 from typing import Optional
 
 
-def get_current_window_linux() -> Optional[dict]:
-    from . import xlib
-    window = xlib.get_current_window()
+class Linux:
+    def __init__(self):
+        try:
+            import pydbus
+            self.bus = pydbus.SessionBus()
+        except ModuleNotFoundError:
+            self.bus = False
+            self.gnome_shell = None
+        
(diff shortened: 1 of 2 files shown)


aw_watcher_window/lib.py
```diff
@@ -1,19 +1,77 @@
...
-        name = xlib.get_window_name(window)
+class Linux:
+    def __init__(self):
+        try:
+            import pydbus
+            self.bus = pydbus.SessionBus()
+        except ModuleNotFoundError:
+            logger.info("pydbus not installed, GNOME-Shell Wayland support disabled")
+            self.bus = False
+            self.gnome_shell = None
```
johan-bjareholt, line 17: self.bus = None and self.gnome_shell = False would make more sense.

aw_watcher_window/lib.py
```diff
@@ -1,19 +1,77 @@
...
+
+    def get_current_window(self) -> dict:
+        if self.gnome_shell:
+            return self.get_current_window_gnome_shell()
+        
+        return self.get_current_window_x11()
+    
+    def _setup_gnome(self) -> None:
+        js_code = """
+        global._aw_current_window = () => {
```
johan-bjareholt, line 35: Nice!
Much better solution now, no race conditions so seems like it should be reliable.

aw_watcher_window/lib.py
```diff
@@ -1,19 +1,77 @@
...
-    from . import xlib
-    window = xlib.get_current_window()
 
-    if window is None:
-        cls = "unknown"
-        name = "unknown"
-    else:
-        cls = xlib.get_window_class(window)
-        name = xlib.get_window_name(window)
+class Linux:
```
johan-bjareholt, line 9: All other platform specific changes are in seperate files so would make more sense to create a new gnome_wayland.py file and only check for which backend to use in lib.py

johan-bjareholt (commented):
Cool that you got this working!

Won't be complete without contributing similar changes to aw-watcher-afk, both sway and kwin implements the idle.xml protocol but I'm not sure if Gnome does that.

Not sure how well pydbus would work with pyinstaller due to it requiring system libraries, will have to look into that. Another option would be to only support Gnome when running/installing from source.

I'd prefer to avoid compositor specific code in aw-watcher-window, but since wayland is the way it is with fragmenting APIs for the linux desktop due to compositor developers making their own wayland APIs instead of working on having common wayland protocols I'm still not against possibly merging something like this in the future.
Having it as a gnome extension though would of course also be an option, but then the installation experience wouldn't be as good if you have to install two things instead of just one I believe.

Jakuje:
Is there any progress on this? I would be glad to start using this on GNOME wayland as I getting tired of useless reports from rescuetime (where I do not expect this will be ever implemented).

johan-bjareholt:
@jakuje Unfortunately there's no progress as you can see. 

If it important for you to use Wayland it is possible to use aw-watcher-window-wayland but that does not work on gnome due to it not implementing the wlr-foreign-toplevel-management Wayland protocol.

Otherwise the status is the same as before, we'd like to avoid merging this if possible because it's gnome specific rather than Linux/Wayland specific, but if we have no option in the future we might merge it anyway.

rickdub2011:
You

rickdub2011 (approved):

jacksongoode:
Any further thoughts on this request?

jopereira:
This no longer works with GNOME 41 that restricts `Eval`. There is a workaround (https://github.com/ramottamado/eval-gjs) but packaging `_aw_current_window` as a Shell extension seems to be now the best alternative.

sgraf812:
FYI, https://extensions.gnome.org/extension/4974/window-calls-extended/ might already provide the info you need

benmoussatMouad:
Not sure if this is relevant or not but it's working perfectly after some work-arounds. 

Since GNOME 41, the `Eval` function no longer works. There is a work around for this : https://askubuntu.com/questions/1412130/dbus-calls-to-gnome-shell-dont-work-under-ubuntu-22-04 

Disabling the unsafe-mode in gnome seems to work perfectly. 

Also running aw-watcher-afk with sudo previlege. No problems so far.

OS: Fedora Linux 37.
DE: GNOME 43 on wayland.

JohannesBuchner:
Thanks @sgraf812, based on the information here I was able to put together a primitive activity logging shell script for gnome+wayland, you can see it at https://github.com/JohannesBuchner/activitytracker/blob/master/track-gnome.sh

mathieujobin:
Thanks @JohannesBuchner 

its sad this never got merged/released...

ErikBjare:
@mathieujobin awatcher (https://github.com/2e3s/awatcher) supports GNOME Wayland if used with [this gnome extension](https://extensions.gnome.org/extension/5592/focused-window-d-bus/)

mathieujobin:
@ErikBjare 

Oh thank you, i am going to try it.

PR #49, opened 2021-02-01
Update macOS window title logic

ErikBjare:
Continuing on #40

### Issues

 - [ ] It seems like screen capture permissions are required in macOS 10.15+ (Catalina) for window titles to show up.
    - See this issue for an example of a fix in Swift: https://github.com/lwouis/alt-tab-macos/issues/127
    - There is a function [CGRequestScreenCaptureAccess](https://developer.apple.com/documentation/coregraphics/3656524-cgrequestscreencaptureaccess) but I can't seem to access it from pyobjc. 
        - **Edit:** Looks like it wasn't added until macOS 11... (https://twitter.com/hrumarek/status/1345593659132051456)
 - [ ] Need to rewrite parts so that it correctly handles the event-based approach.
 - [ ] Need to combine event-based approach with polling since there's no event for 'window title changed' (only 'application changed').
 - [ ] The Python docs state that multiprocessing `set_start_method`s 'spawn' and 'forkserver' cannot be used with 'frozen' executables (as produced by PyInstaller). But 'spawn' became the default in Python 3.8, and the Python 3.7 default of 'fork' seems to make the new process crash...
 - [ ] More stuff from/comments on #40?


ErikBjare:
Hey @xylix I made some progress here and seem to have tracked down the issue as to why window titles aren't being returned (screen recording permissions needed...)

Would you be interested in (and have the time) to continue work on this in the near future? I don't think it should be that much work now that I found the core issue.

xylix:
I might have time in a couple weeks... can't really promise I will, though.  It is probably pretty doable in a few hours of work at this point.

On the fork / subprocess stuff: I've coded some other python multiprocessing and came across fork crashing things on macOS. Only being able to use fork with pyinstaller seems like a quite difficult limitation to work around 🤔

ErikBjare:
Understandable, just checking :)

tbh I think it might sense to make a release supporting both the new and old method (with the old as default, for the time being), so we can get something basic merged and go from there.

ErikBjare:
@iloveitaly Basically everything mentioned in the PR description.

I've abandoned work on this for now, as you provided an alternative method that wasn't as cumbersome to implement (but may still suffer from some of the same bugs).

iloveitaly:
@ErikBjare any notes around what's left here to do?

ErikBjare:
@iloveitaly Only what's in the PR description I think.

iloveitaly:
@ErikBjare we should close this out too—the swift-based method solves all of these problems!

TimeToBuildBob:
## 🤖 AI code review

This PR rewrites the macOS window-title capture from an osascript subprocess call to an event-driven AppKit/Quartz implementation. It adds a new init() entry point in lib.py that hands the main thread to a Cocoa event loop and runs the heartbeat callback on a daemon thread, replaces the old getInfo/getApp/getTitle functions with a Watcher/Observer queue, and removes the bundled printAppTitle.scpt from the repo and the PyInstaller spec. It also updates the README heading and adds a TODO in main.py about not sleeping on macOS.

### Not safe to merge — 4 P1 open

Confidence 2/5

<details><summary><b>4 findings</b> · ❌ 4 P1 · 🔒 1 security</summary>

**❌ P1 high** — `aw_watcher_window/macos.py:138`

In background_ensure_permissions(), the code unconditionally starts a Process(target=show_screencapture_permissions_dialog) even when accessibility permissions are already granted, and it does so regardless of macOS version. The comment says screen capture permissions are needed on macOS 10.15+, but the code does not check the OS version, so on macOS 10.14 and earlier the dialog is shown to every user on every startup even though screen capture permission is not required. Additionally, the dialog is shown even when the user already has accessibility permissions, so a user who has granted the required permission still gets a modal 'Missing screen capture permissions' alert. The observable consequence is that macOS users are repeatedly prompted with a misleading permission dialog, and on older macOS versions the dialog is entirely spurious. The code also calls set_start_method('spawn') unconditionally; if this module is imported more than once in the same process, set_start_method raises RuntimeError, but that is secondary.

```suggestion
if not accessibility_permissions:
        Process(target=show_screencapture_permissions_dialog, args=(())).start()
```

<sub>**How this was verified:** Checked the function body: lines 125-138 show the if/else only prints, then line 138 unconditionally starts the process. The PR description also notes screen capture permissions are required on 10.15+, but the code never checks them before showing the dialog.</sub>

**❌ P1 high** — `aw_watcher_window/macos.py:81`

The new event-based macOS flow never emits an initial event until the user switches applications. Observer._set_front_app() is only called from handle_, which is triggered by NSWorkspaceDidActivateApplicationNotification. On startup, no such notification fires, so the queue is empty and get_window_event() blocks forever in queue.get(). The heartbeat loop in main.py calls get_current_window() on the daemon thread, which blocks indefinitely, so no heartbeat is ever sent until the user switches apps. Even after the first switch, the loop sleeps poll_time after each event, so the watcher only reports the app that was active at the moment of the switch, not the current window title at poll time. The PR description itself notes that there is no event for window title changes, but the code does not combine polling with the event queue as the TODO suggests. The observable consequence is that on macOS the watcher produces no data at startup and stale data between app switches.

```suggestion
def __init__(self):
        self.observer = Observer.new()
        self.observer._set_front_app()  # enqueue initial event
```

<sub>**How this was verified:** Traced the flow: init() -&gt; hand_over_main -&gt; callback thread runs heartbeat_loop; get_current_window -&gt; get_window_event -&gt; watcher.get_next_event -&gt; queue.get() with no timeout. The only producer is handle_ which requires a notification. No initial event is enqueued in Watcher.__init__ or init().</sub>

**❌ P1 high · 🔒 security** — `aw_watcher_window/macos.py:98`

The new macOS implementation uses `print()` statements for logging (e.g., 'Initializing...', 'Event received', 'Checking if we have accessibility permissions...') instead of the logger. In a packaged PyInstaller app with console=True, these prints go to stdout, but in a normal daemonized run they may be lost or interleaved. More importantly, the `print(AXIsProcessTrusted())` and `print(window)` in get_app_title() will print the entire window dictionary, which may contain sensitive window titles, to stdout on every event. This is a privacy leak: window titles can contain sensitive information (e.g., document names, URLs, chat messages), and printing them to the console/logs exposes that data to anyone who can read the process output. The PR description does not mention any logging of window titles, and the previous implementation did not print them. The observable consequence is that sensitive window titles are written to stdout, potentially ending up in log files or terminal output.

```suggestion
logger.debug(window)
```

<sub>**How this was verified:** Lines 43, 97-98, 125, 130 use print. The window dict from CGWindowListCopyWindowInfo contains kCGWindowName (title) and kCGWindowOwnerPID. The PR description also mentions print statements as an issue.</sub>

**❌ P1 high** — `aw_watcher_window/lib.py:23`

In lib.py, the return type annotations for get_current_window_linux, get_current_window_macos, and get_current_window_windows are changed from Optional[dict] to Dict[str, str], but the functions can still return None in some paths? Actually get_current_window_linux always returns a dict, get_current_window_macos returns the event dict, get_current_window_windows returns a dict. However, get_current_window() can raise an exception for unknown platforms, which is fine. But the type change is a contract change for any external callers that relied on Optional. The PR description does not mention this as a breaking change. However, the more concrete issue is that get_current_window_macos now returns the event's dict which has keys 'app' and 'title', while the other platforms return 'appname' and 'title'. In heartbeat_loop, the code accesses current_window['appname'] (line 81). On macOS, the dict from get_window_event has key 'app', not 'appname', so this will raise KeyError, which is caught by the except in heartbeat_loop and logs an error, then sets current_window to unknown. This means on macOS, the watcher will never record the actual app name; it will always log 'unknown' for the app. This is a critical bug.

```suggestion
return {"appname": macos.get_window_event()[1]["app"], "title": macos.get_window_event()[1]["title"]}
```

<sub>**How this was verified:** In macos.py line 50, the event dict is {"app": get_app_name(app), "title": get_app_title(app)}. In main.py line 81, data uses current_window["appname"]. The KeyError is caught and replaced with unknown. This is a clear mismatch.</sub>

</details>

<details><summary><b>Files changed (5)</b> — the diff as I read it</summary>

- `README.md` — Changes the top-level heading from an underline-style to an ATX-style markdown heading.
- `aw-watcher-window.spec` — Removes the printAppTitle.scpt data file from the PyInstaller analysis datas list.
- `aw_watcher_window/lib.py` — Adds an init() callback-based entry point, changes macOS window retrieval to use macos.get_window_event(), and tightens return type annotations.
- `aw_watcher_window/macos.py` — Replaces the osascript-based getInfo/getApp/getTitle with an AppKit event-loop Watcher/Observer queue and adds permission-dialog helpers.
- `aw_watcher_window/main.py` — Adds a TODO comment noting that sleeping on macOS may be unnecessary with the event-based approach.

</details>

<sub>Reviewed `9af99932aee1` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 679s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw_watcher_window/macos.py
```diff
@@ -1,45 +1,196 @@
...
+    else:
+        print("We don't, showing dialog.")
+        Process(target=show_screencapture_permissions_dialog, args=(())).start()
+
+    # Needed on macOS 10.15+
+    # TODO: Add macOS version check to ensure it doesn't break on lower versions
+    # This would have been nice, but can't seem to call it through pyobjc... ('unknown location')
+    # from Quartz.CoreGraphics import CGRequestScreenCaptureAccess
+    # screencapture_permissions = CGRequestScreenCaptureAccess()
+    Process(target=show_screencapture_permissions_dialog, args=(())).start()
```
TimeToBuildBob, line 138: ❌ **P1** — In background_ensure_permissions(), the code unconditionally starts a Process(target=show_screencapture_permissions_dialog) even when accessibility permissions are already granted, and it does so regardless of macOS version. The comment says screen capture permissions are needed on macOS 10.15+, but the code does not check the OS version, so on macOS 10.14 and earlier the dialog is shown to every user on every startup even though screen capture permission is not required. Additionally, the dialog is shown even when the user already has accessibility permissions, so a user who has granted the required permission still gets a modal 'Missing screen capture permissions' alert. The observable consequence is that macOS users are repeatedly prompted with a misleading permission dialog, and on older macOS versions the dialog is entirely spurious. The code also calls set_start_method('spawn') unconditionally; if this module is imported more than once in the same process, set_start_method raises RuntimeError, but that is secondary.

aw_watcher_window/macos.py
```diff
@@ -1,45 +1,196 @@
...
+        #     None)
+        try:
+            AppHelper.runConsoleEventLoop()
+        except KeyboardInterrupt:
+            print("Main thread was asked to stop, quitting.")
+
+    def stop(self):
+        AppHelper.stopEventLoop()
+
+    def get_next_event(self) -> Tuple[float, Dict[str, str]]:
```
TimeToBuildBob, line 81: ❌ **P1** — The new event-based macOS flow never emits an initial event until the user switches applications. Observer._set_front_app() is only called from handle_, which is triggered by NSWorkspaceDidActivateApplicationNotification. On startup, no such notification fires, so the queue is empty and get_window_event() blocks forever in queue.get(). The heartbeat loop in main.py calls get_current_window() on the daemon thread, which blocks indefinitely, so no heartbeat is ever sent until the user switches apps. Even after the first switch, the loop sleeps poll_time after each event, so the watcher only reports the app that was active at the moment of the switch, not the current window title at poll time. The PR description itself notes that there is no event for window title changes, but the code does not combine polling with the event queue as the TODO suggests. The observable consequence is that on macOS the watcher produces no data at startup and stale data between app switches.

aw_watcher_window/macos.py
```diff
@@ -1,45 +1,196 @@
...
+def get_app_title(app: NSRunningApplication) -> str:
+    pid = app.processIdentifier()
+    options = kCGWindowListOptionOnScreenOnly
+    windowList = CGWindowListCopyWindowInfo(options, kCGNullWindowID)
+
+    for window in windowList:
+        lookupPid = window['kCGWindowOwnerPID']
+        if (lookupPid == pid):
+            print(AXIsProcessTrusted())
+            print(window)
```
TimeToBuildBob, line 98: ❌ **P1** — The new macOS implementation uses `print()` statements for logging (e.g., 'Initializing...', 'Event received', 'Checking if we have accessibility permissions...') instead of the logger. In a packaged PyInstaller app with console=True, these prints go to stdout, but in a normal daemonized run they may be lost or interleaved. More importantly, the `print(AXIsProcessTrusted())` and `print(window)` in get_app_title() will print the entire window dictionary, which may contain sensitive window titles, to stdout on every event. This is a privacy leak: window titles can contain sensitive information (e.g., document names, URLs, chat messages), and printing them to the console/logs exposes that data to anyone who can read the process output. The PR description does not mention any logging of window titles, and the previous implementation did not print them. The observable consequence is that sensitive window titles are written to stdout, potentially ending up in log files or terminal output.

aw_watcher_window/lib.py
```diff
@@ -16,16 +16,14 @@ def get_current_window_linux() -> Optional[dict]:
...
 
-def get_current_window_macos() -> Optional[dict]:
+def get_current_window_macos() -> Dict[str, str]:
     from . import macos
-    info = macos.getInfo()
-    app = macos.getApp(info)
-    title = macos.getTitle(info)
+    # TODO: This should return the latest event, or block until there is one
+    # TODO: This currently discards the event timestamp, but it should propagate upwards...
+    return macos.get_window_event()[1]
```
TimeToBuildBob, line 23: ❌ **P1** — In lib.py, the return type annotations for get_current_window_linux, get_current_window_macos, and get_current_window_windows are changed from Optional[dict] to Dict[str, str], but the functions can still return None in some paths? Actually get_current_window_linux always returns a dict, get_current_window_macos returns the event dict, get_current_window_windows returns a dict. However, get_current_window() can raise an exception for unknown platforms, which is fine. But the type change is a contract change for any external callers that relied on Optional. The PR description does not mention this as a breaking change. However, the more concrete issue is that get_current_window_macos now returns the event's dict which has keys 'app' and 'title', while the other platforms return 'appname' and 'title'. In heartbeat_loop, the code accesses current_window['appname'] (line 81). On macOS, the dict from get_window_event has key 'app', not 'appname', so this will raise KeyError, which is caught by the except in heartbeat_loop and logs an error, then sets current_window to unknown. This means on macOS, the watcher will never record the actual app name; it will always log 'unknown' for the app. This is a critical bug.

PR #52, opened 2021-05-01
Use JXA for window status on macOS, include url in event data

iloveitaly:
This will allow us to create rules based on domains instead of just the window titles for more accurate tracking.


iloveitaly:
@ErikBjare you look like the best person to review. Take a look and let me know what you think!

ErikBjare:
Thanks for taking the time to work on this!

You're not the first to have made the change to JS, and there has been quite a bit of discussion around this, and why all AppleScript/JS approaches are undesirable from a performance standpoint. See #29 and #40.

Clever that you managed to get the URL from the Accessibility API! However, the planned/preferred platform- and browser-independent way is to use aw-watcher-web to collect URLs, we just need to get them merged into the window events (there's an open PR in aw-server-rust that aims to do just that).

iloveitaly:
Thanks for the quick reply! https://github.com/ActivityWatch/aw-watcher-window/pull/49 looks interesting—the event-based approach is definitely an improvement over polling!

> However, the planned/preferred platform- and browser-independent way is to use aw-watcher-web to collect URLs

Curious why this is the preferred approach. For FireFox, it's the *only* approach, but for the all Chrome-based browsers and Safari the jxa approach is nearly the same (you could `execute` javascript on the tab to extract additional properties if needed).

From a UX perspective, extensions are a worse experience. Users would need to install a separate browser extension, on each browser they use. If they use Brave, Chrome, and Google Chrome Canary they'd need to install a separate browser on each. Then, each of those extensions would send their individual heartbeats to aw-server, which adds more load to the system (more heartbeats). Additionally, updates to the extension wouldn't be deployed at the same time as updates to the core AW software.

I think for FF, an extension is definitely needed, but I wonder if we could keep things more simple and use JXA for all of the other browsers.

> and why all AppleScript/JS approaches are undesirable from a performance standpoint

A JXA approach definitely won't be as performant as a pyobjc bridge-based solution, but (a) if we are using eventing vs polling and (b) if we [executed JXA directly](https://github.com/Hammerspoon/hammerspoon/blob/da2f41aa2c008c84a746ffc0d5a0d3291d6e5264/extensions/osascript/internal.m#L24) vs executing via `osascript` my guess is the performance his would be negligible, especially when compared to heartbeats running across 2+ browsers on a user's machine. 

What do you think? Thanks for being open to a conversation here!

ErikBjare:
> Curious why this is the preferred approach.

Because its the only solution on other platforms than macOS.

From a dev perspective it's important that things work roughly the same way across platforms, to minimize special/platform-specific cases.

I'm still open to merging something like this and having the URL watching enabled either by default or through a flag/config option, but care needs to be taken so that it doesn't conflict with the web extension approach. 

> Additionally, updates to the extension wouldn't be deployed at the same time as updates to the core AW software.

I don't think this is an issue. WebExtensions generally update automatically, and breaking changes in the heartbeat API should be extremely rare.

> (performance comments)

If you can make the AppleScript/JXA more performant by calling it directly instead of spawning a new process, that'd be great! 

But in the end the AppleScript approach (and I assume the same holds for JXA) has been a frequent cause of issues/bugs (often relating to permissions). It makes sense that such is the case, as it seems the AppleScript approach gives us permissions we otherwise shouldn't have (as seen in #49 where I need *more* permissions than required by the AppleScript approach).

> ...especially when compared to heartbeats running across 2+ browsers on a user's machine.

The web watcher doesn't use polling, so the number of browsers used shouldn't impact performance.

ErikBjare:
And as I mentioned in #49, I'm open to having multiple approaches merged and having configuration options for which specific one to use.

This would help evaluate the different ways to go about this, without getting bogged down by choosing a particular approach.

iloveitaly:
@ErikBjare rewrote this to avoid shelling out to osascript for execution. Going to let this bake locally for a bit to see if any issues come up. 

This feels like a improvement over the existing implementation and will only run on macOS, so I'm not sure we need a flag to enable this right now, although that may make sense when work on the other PR is done.

iloveitaly:
Another benefit to this implementation which I just noticed is Electron-based applications report their real name, instead of simply "Electron"

aw_watcher_window/lib.py
```diff
@@ -13,16 +13,13 @@ def get_current_window_linux() -> Optional[dict]:
         cls = xlib.get_window_class(window)
         name = xlib.get_window_name(window)
 
-    return {"appname": cls, "title": name}
+    return {"app": cls, "title": name}
 
 
 def get_current_window_macos() -> Optional[dict]:
+    # TODO should we use unknown when the title is blank like the other platforms?
```
ErikBjare, line 20: I think so yes, I'll elaborate in #54.

ErikBjare (commented):
Looks good! 

I want to merge this, but in case there are any unexpected side effects (on old/new macOS versions or such) I'd like to keep the old approach in the codebase. 

Would you mind moving the JXA approach to a new file and restore the old one? Then add a config option to let users set which method to use.

And again: Great work! Looking forward to our call tomorrow :)

ErikBjare:
@iloveitaly Just a friendly ping that I'd really like to get this merged. 

If you could make the requested change I'll merge it asap :)

iloveitaly:
@ErikBjare updated! Let me know what you think

aw_watcher_window/main.py
```diff
@@ -45,43 +43,31 @@ def main():
 
     sleep(1)  # wait for server to start
     with client:
-        heartbeat_loop(client, bucket_id, poll_time=args.poll_time, exclude_title=args.exclude_title)
-
+        heartbeat_loop(client, bucket_id,
+        poll_time=args.poll_time, exclude_title=args.exclude_title, macos_strategy=args.macos_strategy)
```
ErikBjare, line 47:
```suggestion
        heartbeat_loop(client, bucket_id, poll_time=args.poll_time, exclude_title=args.exclude_title, strategy=args.strategy)
```

aw_watcher_window/main.py
```diff
@@ -45,43 +43,31 @@ def main():
...
-    parser = argparse.ArgumentParser("A cross platform window watcher for Activitywatch.\nSupported on: Linux (X11), macOS and Windows.")
-    parser.add_argument("--testing", dest="testing", action="store_true")
-    parser.add_argument("--exclude-title", dest="exclude_title", action="store_true", default=default_exclude_title)
-    parser.add_argument("--verbose", dest="verbose", action="store_true")
-    parser.add_argument("--poll-time", dest="poll_time", type=float, default=default_poll_time)
-    return parser.parse_args()
-
-
-def heartbeat_loop(client, bucket_id, poll_time, exclude_title=False):
+def heartbeat_loop(client, bucket_id, poll_time, macos_strategy, exclude_title=False):
```
ErikBjare, line 49:
```suggestion
def heartbeat_loop(client, bucket_id, poll_time, strategy, exclude_title=False):
```

aw_watcher_window/main.py
```diff
@@ -45,43 +43,31 @@ def main():
...
     while True:
         if os.getppid() == 1:
             logger.info("window-watcher stopped because parent process died")
             break
 
         try:
-            current_window = get_current_window()
+            current_window = get_current_window(macos_strategy=macos_strategy)
```
ErikBjare, line 56:
```suggestion
            current_window = get_current_window(strategy)
```

aw_watcher_window/lib.py
```diff
@@ -13,16 +12,19 @@ def get_current_window_linux() -> Optional[dict]:
...
 
-    return {"title": title, "appname": app}
+    # `jxa` is the default & preferred strategy. It includes the url + incognito status
+    if strategy == 'jxa':
+        from . import macos
+        return macos.getInfo()
+    else:
+        from . import macos_applescript
+        return macos_applescript.getInfo()
```
ErikBjare, line 27:
```suggestion
    elif strategy == 'applescript':
        from . import macos_applescript
        return macos_applescript.getInfo()
    else:
        return ValueError(f"invalid strategy '{strategy}'")
```

aw_watcher_window/config.py
```diff
@@ -1,18 +1,31 @@
...
-    default_client_config["aw-watcher-window"] = {
-        "exclude_title": False,
-        "poll_time": "1.0"
-    }
-    default_client_config["aw-watcher-window-testing"] = {
+    default_client_config["aw-watcher-window"] = default_client_config["aw-watcher-window-testing"] = {
         "exclude_title": False,
-        "poll_time": "1.0"
+        "poll_time": "1.0",
+        "macos_strategy": "jxa"
```
ErikBjare, line 11:
```suggestion
        "strategy": "jxa"
```

aw_watcher_window/config.py
```diff
@@ -1,18 +1,31 @@
...
+    default_poll_time = config.getfloat("poll_time")
+    default_exclude_title = config.getboolean("exclude_title")
+    default_macos_strategy = config.get("macos_strategy")
+
+    parser = argparse.ArgumentParser("A cross platform window watcher for Activitywatch.\nSupported on: Linux (X11), macOS and Windows.")
+    parser.add_argument("--testing", dest="testing", action="store_true")
+    parser.add_argument("--exclude-title", dest="exclude_title", action="store_true", default=default_exclude_title)
+    parser.add_argument("--verbose", dest="verbose", action="store_true")
+    parser.add_argument("--poll-time", dest="poll_time", type=float, default=default_poll_time)
+    parser.add_argument("--macos-strategy", dest="macos_strategy", default=default_macos_strategy, choices=["jxa", "applescript"])
```
ErikBjare, line 29:
```suggestion
    parser.add_argument("--strategy", dest="strategy", default=default_strategy, choices=["jxa", "applescript"], help="(macOS only) strategy to use for retrieving the active window")
```

aw_watcher_window/config.py
ErikBjare: Did this change work without you having to manually edit the config file? (we've had some issues where new defaults won't be applied for old installs...)

ErikBjare (commented):
Looks good! 

A few nits, mostly regarding variable naming. Also:

 - [ ] Need to restore the `printAppTitle.scpt` file

aw_watcher_window/config.py
ErikBjare: I now realize why you chose to prepend `macos_` to the strategy...

aw_watcher_window/config.py
```diff
@@ -1,18 +1,31 @@
...
 
     # TODO: Handle so aw-watcher-window testing gets loaded instead of testing is on
     return _load_config("aw-watcher-window", default_client_config)["aw-watcher-window"]
+
+def parse_args():
+    config = load_config()
+
+    default_poll_time = config.getfloat("poll_time")
+    default_exclude_title = config.getboolean("exclude_title")
+    default_macos_strategy = config.get("macos_strategy")
```
ErikBjare, line 22:
```suggestion
    default_strategy_macos = config.get("strategy_macos")
```

aw_watcher_window/main.py
```diff
@@ -11,22 +10,21 @@
...
-    config = load_config()
-    args = parse_args(
-        default_poll_time=config.getfloat("poll_time"),
-        default_exclude_title=config.getboolean("exclude_title"),
-    )
-
     if sys.platform.startswith("linux") and ("DISPLAY" not in os.environ or not os.environ["DISPLAY"]):
         raise Exception("DISPLAY environment variable not set")
 
+    args = parse_args()
```
ErikBjare, line 26: This should happen before the check above in order for the CLI to behave properly in the absence of an X server.

aw_watcher_window/config.py
iloveitaly: I'm confused here: do you want the var name to change before merging?

aw_watcher_window/main.py
iloveitaly: Ah, got it. Will move!

iloveitaly:
> * Need to restore the `printAppTitle.scpt` file

This was restored, although it it modified from the original file. The source code is exactly the same, so I'm guessing the difference in the file was caused by the more recent macos version I'm on.

I've restored the original version in the latest push. Take a look and let me know if there are any other tweaks!

aw_watcher_window/config.py
ErikBjare: No sorry, I already applied the suggested change :)

ErikBjare:
Looks good! Waiting for the CI to pass, and then I'll merge :)

ErikBjare:
Looks like type checking fails in a few places. I'll clean it up.

ErikBjare:
Merged! 

Thanks for all your work on this! Hopefully it'll lead to a lot more reliable macOS experience in the next release :)

PR #53, opened 2021-05-04
build(deps): added pyinstaller to optional dependencies

iloveitaly:
This was required in order to run `make package`. 

Also bumped pyobjc versions. This eliminates this error I ran into `RuntimeError: Wrong version of PyObjC C API (got 21, expected 20)`.


pyproject.toml
```diff
@@ -14,15 +14,16 @@ aw-client = {git = "https://github.com/ActivityWatch/aw-client.git"}
...
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.0"
 mypy = "*"
 macholib = {version = "^1.13", platform = "darwin"}  # Needed for pyinstaller
+pyinstaller = "^4.3"
 
 [build-system]
-requires = ["poetry>=0.12"]
+requires = ["poetry_core>=0.12"]
```
iloveitaly, line 28: in the poetry docs, it was recommended to reference `poetry_core` instead in the build-system section.

johan-bjareholt:
PyInstaller is only an optional dependency so it should not be in the dependencies.

Could you change the PR to only bump up the pyobjc dependencies?

iloveitaly:
Is it optional because someone could build & run from the cli without packing it up within an application?

> Could you change the PR to only bump up the pyobjc dependencies?

What do you think about `optional = true` instead of removing it, since it's required to run `make package`?

ErikBjare:
>Is it optional because someone could build & run from the cli without packing it up within an application?

It's optional because when we bundle everything in the main activitywatch repo it gets to decide which pyinstaller version to use (as it should use the same version for all modules).

>What do you think about optional = true instead of removing it, since it's required to run make package?

That'd be alright! As long as it isn't installed by default (which would interfere with pre-installed versions of pyinstaller, as in the main/bundle repo's CI).

PR #54, opened 2021-05-06
fix: don't log "unknown" event if polling failed

iloveitaly:
I don't think we should emit a 'unknown' window/app event if get_current_window() fails. If the function returns `None` than we omit an event, so not logging an unknown event on an exception being raised will be more consistent. Additionally, these unknown events create noise in activity reporting.

﻿logging.exception was used to simplify exception logging:
https://stackoverflow.com/questions/1508467/log-exception-with-traceback

### DIFF
--- a/aw_watcher_window/main.py
+++ b/aw_watcher_window/main.py
@@ -66,6 +66,8 @@ def heartbeat_loop(client, bucket_id, poll_time, strategy, exclude_title=False):
             logger.info("window-watcher stopped because parent process died")
             break
 
+        current_window = None
+
         try:
             current_window = get_current_window(strategy)
             logger.debug(current_window)
--- a/aw_watcher_window/main.py
+++ b/aw_watcher_window/main.py
@@ -1,5 +1,4 @@
 import logging
-import traceback
 import sys
 import os
 from time import sleep
--- a/aw_watcher_window/main.py
+++ b/aw_watcher_window/main.py
@@ -71,11 +71,7 @@ def heartbeat_loop(client, bucket_id, poll_time, strategy, exclude_title=False):
             current_window = get_current_window(strategy)
             logger.debug(current_window)
         except Exception as e:
-            logger.error(
-                "Exception thrown while trying to get active window: {}".format(e)
-            )
-            traceback.print_exc()
-            current_window = {"app": "unknown", "title": "unknown"}
+            logger.exception("Exception thrown while trying to get active window")
 
         now = datetime.now(timezone.utc)
         if current_window is None:


ErikBjare:
The reason we log "unknown" events is to make it clear to the user that there wasn't an "active" app to get (as is the case on some platforms when you're just on your desktop) or we otherwise couldn't get the title/it was missing.

>If the function returns None than we omit an event, so not logging an unknown event on an exception being raised will be more consistent.

I did not realize that!

I also notice that the failure message only has a log level of debug, which would make the error not show up in logs by default either.

https://github.com/ActivityWatch/aw-watcher-window/blob/c9bf8208a9a4aaa6391453da7577c41e55757916/aw_watcher_window/main.py#L76-L77

>Additionally, these unknown events create noise in activity reporting.

Indeed, but I think the solution should preferably be to fix the source of the errors instead of hiding them. 

Not sending an event at all would also create noise in the reporting, but it would be harder to spot as it'd likely get approximated away by flooding at the analysis stage (see https://github.com/ActivityWatch/activitywatch/issues/124).

If you make it emit 'unknown' events under both error conditions, do you get them often?

aw_watcher_window/main.py
```diff
@@ -68,9 +68,7 @@ def heartbeat_loop(client, bucket_id, poll_time, exclude_title=False):
             current_window = get_current_window()
             logger.debug(current_window)
         except Exception as e:
-            logger.error("Exception thrown while trying to get active window: {}".format(e))
-            traceback.print_exc()
-            current_window = {"appname": "unknown", "title": "unknown"}
+            logging.exception("Exception thrown while trying to get active window")
```
ErikBjare, line 71: This is good, but should use the module logger.

```suggestion
            logger.exception("Exception thrown while trying to get active window")
```

iloveitaly:
> Indeed, but I think the solution should preferably be to fix the source of the errors instead of hiding them.

Agreed, although, at least on macos this may not always be possible. In my experience with the applescript/jxa approach, there's intermittent errors that occur every once in a while due to closing a window right when it's being inspected, etc. The only way around this error from my perspective is:

* possibly the non-applescript based code in another PR (I'm doubtful this would fix the issue completely)
* backoff retries
* catching specific errors within the os-specific logic and returning None

Backoff doesn't make sense given the script runs every 1s. Best approach here in my mind is to try/catch the exception in the underlying os-specific code and set `current_window = {"appname": "unknown", "title": "unknown"}` if an exception is raised. Wdyt?

iloveitaly:
@ErikBjare friendly reminder on this PR!

aw_watcher_window/main.py
```diff
@@ -71,11 +70,7 @@ def heartbeat_loop(client, bucket_id, poll_time, strategy, exclude_title=False):
             current_window = get_current_window(strategy)
             logger.debug(current_window)
         except Exception as e:
-            logger.error(
-                "Exception thrown while trying to get active window: {}".format(e)
-            )
-            traceback.print_exc()
-            current_window = {"app": "unknown", "title": "unknown"}
+            logger.exception("Exception thrown while trying to get active window")
```
ErikBjare, line 73: This will incorrectly send the last `current_window` as the new `current_window`. Needs a `sleep()` and `continue` to prevent it.

iloveitaly: 🤦 fixed!

ErikBjare:
Nice, merging now.

Thanks again!

PR #59, opened 2021-09-16
fix: fixed jxa strategy where script fails compilation due to shebang line

### LINKED ISSUE #58: [macOS] Exception thrown while trying to get active window: error compiling jxa script
On a fresh installation of ActivityWatch on Mac, the `aw-watcher-window` application failed to work, resulting in all applications reporting as "unknown".

image

 Inspecting the logs, this was the error message, logged approximately every second: 

```
[ERROR]: Exception thrown while trying to get active window: error compiling jxa script  (aw_watcher_window.main:74)
```

When I pasted `printAppStatus.jxa` into the script editor to debug, the error message reported 

```
Error on line 1: SyntaxError: Invalid character: '#'
```

Opening the package contents and deleting line 1 from `printAppStatus.jxa` then solved the problem for me.

I didn't see this error come up in any google searches or existing issues, which was odd to me because the first thing I tried was to remove and reinstall the app a couple times but this bug was consistent on a fresh install on my machine (running Mojave 10.14.6). It's a simple code fix, as seen above, but not seeing the bug reported elsewhere, I wasn't sure if it was affecting anybody else's machine. It seems like it should have, given that the bug was a syntax error in the jxa file.

ErikBjare:
Should fix #58 (untested)

### DIFF
--- a/aw_watcher_window/macos_jxa.py
+++ b/aw_watcher_window/macos_jxa.py
@@ -8,35 +8,47 @@ script = None
 
 
 def compileScript():
-    # https://stackoverflow.com/questions/44209057/how-can-i-run-jxa-from-swift
-    # https://stackoverflow.com/questions/16065162/calling-applescript-from-python-without-using-osascript-or-appscript
+    """
+    Compiles the JXA script and caches the result.
+
+    Resources:
+     - https://stackoverflow.com/questions/44209057/how-can-i-run-jxa-from-swift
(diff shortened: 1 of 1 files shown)


ErikBjare:
@iloveitaly Would be great if you could take a look at this and test it (I don't have access to a macOS VM right now)

ErikBjare:
Going to go ahead and eagerly merge this. Any issues will hopefully surface in pre-release testing.

iloveitaly:
Interesting, I didn't run into issues with the previous config. Just pulled this and running it locally—I'll report any issues.

PR #64, opened 2022-02-07
Fixes "unknown" being returned when some apps open

michaeljelly:
I've been developing an electron application, and noticed that ActivityWatch regularly logged the window as unknown. From research, it seems that some applications aren't "scriptable" according to JXA/applescript, and so you can't get their attributes or properties, which was leading the original script to fail at executing both when setting mainWindow and later when setting title. This meant that everything for that window was recorded as unknown.

To fix this, I first try to get the window the old way, and if this fails then to get the name of the frontmost window as the title. This works very well, it seems to get appName and title accurately, and avoids failing to execute and getting Unknown app names.

Hope this helps!

### DIFF
--- a/aw_watcher_window/printAppStatus.jxa
+++ b/aw_watcher_window/printAppStatus.jxa
@@ -24,7 +24,7 @@ switch(appName) {
   case "Safari":
     // incognito is not available via safari applescript
     url = Application(appName).documents[0].url();
-	  title = Application(appName).documents[0].name();
+    title = Application(appName).documents[0].name();
     break;
   case "Google Chrome": 
   case "Google Chrome Canary":
@@ -36,6 +36,10 @@ switch(appName) {
     url = activeTab.url();
     title = activeTab.name();
     incognito = activeWindow.mode() === 'incognito';
+    break;
+  case "Firefox":
+  case "Firefox Developer Edition":
+    title = Application(appName).windows[0].name();
     break;
   default:
     try {
--- a/aw_watcher_window/printAppStatus.jxa
+++ b/aw_watcher_window/printAppStatus.jxa
@@ -24,9 +24,9 @@ switch(appName) {
   case "Safari":
     // incognito is not available via safari applescript
     url = Application(appName).documents[0].url();
-    title = Application(appName).documents[0].name();
+	  title = Application(appName).documents[0].name();
     break;
-  case "Google Chrome":
+  case "Google Chrome": 
   case "Google Chrome Canary":
   case "Chromium":
   case "Brave Browser":
@@ -37,23 +37,25 @@ switch(appName) {
     title = activeTab.name();
     incognito = activeWindow.mode() === 'incognito';
     break;
-  case "Firefox":
-  case "Firefox Developer Edition":
-    title = Application(appName).windows[0].name();
-    break;
   default:
-    mainWindow = oProcess.
-      windows().
-      find(w => w.attributes.byName("AXMain").value() === true)
+    try {
+      mainWindow = oProcess.windows().find((w, i) => w && w.attributes && w.attributes() && w.attributes.byName("AXMain") && w.attributes.byName("AXMain").value() === true)
+    } catch {
+        mainWindow = oProcess.windows()[0]
+      };
+    
 
     // in some cases, the primary window of an application may not be found
     // this occurs rarely and seems to be triggered by switching to a different application
-    if(mainWindow) {
-      title = mainWindow.
-        attributes.
+    try {
+      title = mainWindow && mainWindow.
+        attributes && mainWindow.attributes.
         byName("AXTitle").
         value()
-    }
+    } catch {
+        title = oProcess.windows[0].name()
+      };
+
 }
 
 // key names must match expected names in lib.py


aw_watcher_window/printAppStatus.jxa
```diff
@@ -42,18 +42,24 @@ switch(appName) {
...
-      find(w => w.attributes.byName("AXMain").value() === true)
+    try {
+      mainWindow = oProcess.windows().find((w, i) => w && w.attributes && w.attributes() && w.attributes.byName("AXMain") && w.attributes.byName("AXMain").value() === true)
+    } catch {
+        mainWindow = oProcess.windows()[0]
+      };
+    
 
     // in some cases, the primary window of an application may not be found
     // this occurs rarely and seems to be triggered by switching to a different application
```
ErikBjare, line 53: It would be great if you could throw in the explanation you have in the PR description as a code comment (optionally with a link to the PR), so that people who read it in the future can understand the context/reason :)

aw_watcher_window/printAppStatus.jxa
```diff
@@ -42,18 +42,24 @@ switch(appName) {
...
     break;
   default:
-    mainWindow = oProcess.
-      windows().
-      find(w => w.attributes.byName("AXMain").value() === true)
+    try {
+      mainWindow = oProcess.windows().find((w, i) => w && w.attributes && w.attributes() && w.attributes.byName("AXMain") && w.attributes.byName("AXMain").value() === true)
+    } catch {
+        mainWindow = oProcess.windows()[0]
+      };
```
ErikBjare, line 49:
```suggestion
    };
```

aw_watcher_window/printAppStatus.jxa
```diff
@@ -42,18 +42,24 @@ switch(appName) {
...
-        attributes.
+    try {
+      title = mainWindow && mainWindow.
+        attributes && mainWindow.attributes.
         byName("AXTitle").
         value()
-    }
+    } catch {
+        title = oProcess.windows[0].name()
+      };
```
ErikBjare, line 61:
```suggestion
    };
```

aw_watcher_window/printAppStatus.jxa
```diff
@@ -42,18 +42,30 @@ switch(appName) {
...
-        attributes.
+    // NEW: I believe the above issue is fixed by the new try-catch structure below which falls back to the name of the frontmost window as the title.
+    try {
+      title = mainWindow && mainWindow.
+        attributes && mainWindow.attributes.
         byName("AXTitle").
         value()
-    }
+    } catch {
+        title = oProcess.windows[0].name()
```
ErikBjare, line 66:
```suggestion
      title = oProcess.windows[0].name()
```

aw_watcher_window/printAppStatus.jxa
```diff
@@ -42,18 +42,30 @@ switch(appName) {
...
-      find(w => w.attributes.byName("AXMain").value() === true)
+    // Some applications aren't "scriptable" according to JXA/applescript, and so you can't get their attributes or properties.
+    // This was leading the original script to fail at executing, both when setting mainWindow and later when setting title.
+    // This meant that everything for that window was recorded as unknown.
+    // To fix this, I first try to get the window the old way, and if this fails then to get the name of the frontmost window as the title.
+    // This works very well, it seems to get appName and title accurately, and avoids failing to execute and getting Unknown app names.
+    try {
+      mainWindow = oProcess.windows().find((w, i) => w && w.attributes && w.attributes() && w.attributes.byName("AXMain") && w.attributes.byName("AXMain").value() === true)
+    } catch {
+        mainWindow = oProcess.windows()[0]
```
ErikBjare, line 53:
```suggestion
      mainWindow = oProcess.windows()[0]
```

michaeljelly:
The code has become slightly uglier to account for Applications where there are no open windows. Let me know whether you prefer just one comment to explain the new try-catch, or a comment in each place.

aw_watcher_window/printAppStatus.jxa
```diff
@@ -23,37 +23,73 @@ var url = undefined, incognito = undefined, title = undefined;
 switch(appName) {
   case "Safari":
     // incognito is not available via safari applescript
+    try { // see catch for details
     url = Application(appName).documents[0].url();
-    title = Application(appName).documents[0].name();
+	  title = Application(appName).documents[0].name();
```
ErikBjare, line 28: Please try to keep whitespace consistent (2 spaces for indentation, no end-of-line spaces)

aw_watcher_window/printAppStatus.jxa
```diff
@@ -23,37 +23,73 @@ var url = undefined, incognito = undefined, title = undefined;
...
+    } catch {
+      // On Mac, you can have an application open without a window.
+      // You can also have safari open with just a Devtools window, meaning no documents
+      // This edge-case previously caused an error meaning time wasn't tracked.
+      // I'm not 100% that a title of "No window" is the right approach, but it is the clearest description of what's happening I can see right now.
+      title = "No window"
```
ErikBjare, line 34: I think a single comment is fine, but perhaps lift out the "No window" constant into a variable, and document the behavior by the declaration/first use?

krlmlr:
This solves the issues I'm having with RStudio on macOS. What's the best way to move forward?

PR #65, opened 2022-05-30
Run macos applescript in fork

iloveitaly:
unfortunately, it's well-documented that applescript leaks memory like crazy:

- https://macscripter.net/viewtopic.php?id=41564
- https://github.com/Hammerspoon/hammerspoon/issues/1980

It's not designed to be executed over and over by the same process. In our case, this
resulted in massive memory usage over time. To work around this, this forks a process
to run our applescript and kill it every so often to clean out it's memory usage.

It's not pretty, but neither is 1GB memory usage for the window watcher. @ErikBjare I'm curious your take on this approach and if I should clean it up for merging.

### DIFF
--- a/aw_watcher_window/__main__.py
+++ b/aw_watcher_window/__main__.py
@@ -3,4 +3,4 @@ from aw_watcher_window import main
 # without this guard, running __main__.py directly will result in a multiprocess
 # related error on macos. This makes it challenging to debug the window watcher on macos
 if __name__ == "__main__":
-  main()
+    main()
--- a/aw_watcher_window/__main__.py
+++ b/aw_watcher_window/__main__.py
@@ -1,3 +1,6 @@
 from aw_watcher_window import main
 
-main()
+# without this guard, running __main__.py directly will result in a multiprocess
+# related error on macos. This makes it challenging to debug the window watcher on macos
+if __name__ == "__main__":
+  main()
--- a/aw_watcher_window/macos_jxa.py
+++ b/aw_watcher_window/macos_jxa.py
@@ -34,9 +34,9 @@ def _compileScript():
     with open(scriptPath) as f:
         scriptContents = f.read()
 
-        # # remove shebang line
-        # if scriptContents.split("\n")[0].startswith("#"):
-        #     scriptContents = "\n".join(scriptContents.split("\n")[1:])
+        # remove shebang line, due to "unexpected token '#'" in some unknown cases/macOS versions"
+        if scriptContents.split("\n")[0].startswith("#"):
+            scriptContents = "\n".join(scriptContents.split("\n")[1:])
 
     script = OSAScript.alloc().initWithSource_language_(
         scriptContents, OSALanguage.languageForName_("JavaScript")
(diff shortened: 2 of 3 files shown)


aw_watcher_window/macos_jxa.py
```diff
@@ -21,17 +23,20 @@ def compileScript():
...
         os.path.dirname(os.path.realpath(__file__)), "printAppStatus.jxa"
     )
+
     with open(scriptPath) as f:
         scriptContents = f.read()
 
-        # remove shebang line
-        if scriptContents.split("\n")[0].startswith("#"):
-            scriptContents = "\n".join(scriptContents.split("\n")[1:])
+        # # remove shebang line
```
iloveitaly, line 37: do you remember why this was needed? I'm not running into any errors with this removed.

aw_watcher_window/__main__.py
```diff
@@ -1,3 +1,6 @@
 from aw_watcher_window import main
 
-main()
+# without this guard, running __main__.py directly will result in a multiprocess
+# related error on macos. This makes it challenging to debug the window watcher on macos
+if __name__ == "__main__":
+  main()
```
ErikBjare, line 6:
```suggestion
    main()
```

aw_watcher_window/macos_jxa.py
```diff
@@ -44,9 +49,109 @@ def compileScript():
 
     return script
 
+def wrapped_function(condition, result_queue: multiprocessing.Queue, function_reference: t.Callable) -> None:
+    # this is run in a forked process, which wipes all logging configuration
```
ErikBjare, line 53: Forking wipes logging config? :O

aw_watcher_window/main.py
```diff
@@ -15,12 +15,6 @@
 
 logger = logging.getLogger(__name__)
 
-# run with LOG_LEVEL=DEBUG
-log_level = os.environ.get("LOG_LEVEL")
-if log_level:
-    logger.setLevel(logging.__getattribute__(log_level.upper()))
-
-
```
ErikBjare, line 23: Why was this removed?

aw_watcher_window/macos_jxa.py
iloveitaly: Seems to. I think it has something to do with the logging config tied to file handles which aren't valid in the forked process? Unsure, was hoping you/someone else understood this more.

aw_watcher_window/main.py
iloveitaly: This is now in aw-client https://github.com/ActivityWatch/aw-client/commit/4b00b98bb39114aedb819e113a7e85f05027024b

iloveitaly:
@ErikBjare I just tried from source. I'll build the app and see it things continue to work.

Question for you: what are your thoughts on increasing poll time to 3s? Polling via applescript takes up a non-trivial amount of CPU time and 1s resolution doesn't get me anything vs 3s (or 10s in my case).

aw_watcher_window/macos_jxa.py
ErikBjare: File handle sounds like a likely culprit. But, just to make sure, "wipes *all* logging configuration" isn't true, is it?

iloveitaly:
@ErikBjare with the new swift-based notifications in place, we won't need this. At some point we should rip out the python-based tracker for macOS as well.

Issue #67, opened 2022-06-18
MacOS: Use realtime accessibility notifications to improve precision and performance


modderme123:
While writing https://github.com/CM-Tech/timeviewer/blob/main/monitor/TimeViewer/AppDelegate.swift, I discovered the swift functions `AXObserverCreate` and `AXObserverAddNotification`. This allows the application to extremely efficiently respond to things like title changes while using almost no CPU (in comparison to the 3% that polling uses on my machine). Would it be possible to use something like this in ActivityWatch to prevent polling?

If this is something that you would be interested in, I'd be open to helping take this on

ErikBjare:
This looks promising! I'd for sure be interested if exploring this. If you can take it on, I'd love that! (and I can make funds available upon request, if that's motivating)

I'd like @iloveitaly's comments on this, as he's been messing in that code the most recently, and has more recent experience with macOS than me :)

iloveitaly:
@modderme123 I've been working on some of the [performance issues here](https://github.com/ActivityWatch/aw-watcher-window/pull/65) (I have it running locally and it's dropped CPU usage quite a bit). 

I think your approach would be way better! Do you think this is something that could be implemented with pyobjc? Or do we need native objc/swift here? I say submit a PR with this change as an alternative (not a replacement) for existing implementation so we can get some folks to test it.

modderme123:
It's likely possible to implement using pyobjc, but I plan to implement it with swift because the above code is already pretty close (once compiled to the command line and converted to heartbeats).

PR #68, opened 2022-07-02
Initial non-polling api

modderme123:
The command to bundle this is `swiftc aw_watcher_window/macos.swift -o aw-window-watcher` (not sure how to add that to the makefile for macos only)

Closes: https://github.com/ActivityWatch/aw-watcher-window/issues/67

### DIFF
--- a/aw_watcher_window/macos.swift
+++ b/aw_watcher_window/macos.swift
@@ -36,9 +36,12 @@ struct Bucket: Codable {
   var hostname: String
 }
 
-let clientHostname = ProcessInfo.processInfo.hostName
-let clientName = "aw-watcher-window2"
-let bucketName = "\(clientName)_\(clientHostname)"
+// Placeholder values, set in start()
+var baseurl = "http://localhost:5600"
+// NOTE: this differs from the hostname we get from Python, here we get `.local`, but in Python we get `.localdomain`
+var clientHostname = ProcessInfo.processInfo.hostName
+var clientName = "aw-watcher-window"
+var bucketName = "\(clientName)_\(clientHostname)"
 
 let main = MainThing()
 var oldHeartbeat: Heartbeat?
@@ -57,6 +60,22 @@ start()
 RunLoop.main.run()
 
 func start() {
+  // Arguments should be:
+  //  - url + port
+  //  - bucket_id
+  //  - hostname
+  //  - client_id
+  let arguments = CommandLine.arguments
+  // Check that we get 4 arguments
+  if arguments.count != 5 {
+    print("Usage: aw-watcher-window <url> <bucket> <hostname> <client>")
+    exit(1)
+  }
+  baseurl = arguments[1]
+  bucketName = arguments[2]
+  clientHostname = arguments[3]
+  clientName = arguments[4]
+
   guard checkAccess() else {
     DispatchQueue.main.asyncAfter(deadline: .now() + 10) {
       start()
@@ -109,7 +128,7 @@ func createBucket() {
   let payload = try! encoder.encode(
     Bucket(client: clientName, type: "currentwindow", hostname: clientHostname))
 
-  let url = URL(string: "http://localhost:5600/api/0/buckets/\(bucketName)")!
+  let url = URL(string: "\(baseurl)/api/0/buckets/\(bucketName)")!
   Task {
     var urlRequest = URLRequest(url: url)
     urlRequest.httpMethod = "POST"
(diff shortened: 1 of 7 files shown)


aw_watcher_window/macos.swift
```diff
@@ -0,0 +1,285 @@
...
+  }
+}
+
+func sendHeartbeat(data: NetworkMessage) {
+  let payload0 = oldHeartbeatData.map {
+    try! encoder.encode(Heartbeat(timestamp: Date.now, data: $0))
+  }
+  let payload = try! encoder.encode(Heartbeat(timestamp: Date.now, data: data))
+
+  let url = URL(string: "http://localhost:5600/api/0/buckets/\(bucketName)/heartbeat?pulsetime=999999")!
```
ErikBjare, line 122: Don't set `pulsetime` to an arbitrarily high value, it will lead to incorrectly merging identical events in case they occur with a long empty interval between (for example, if the computer goes to sleep).

Set it to whatever the expected time is since the last known event (for polling, this is a value slightly higher than the poll interval). 

Since this doesn't use polling, you have to check that the last event was triggered somewhat recently, or combine it with fallback-polling if no event has been triggered recently (for example, in the last minute or so).

The other watchers work in a way such that ActivityWatch should receive an event even if the window title doesn't change (given enough time), adopting a "listen for changes, but poll if long since last check" approach would also help ensure this behavior stays consistent across platforms.

aw_watcher_window/macos.swift
```diff
@@ -0,0 +1,285 @@
...
+
+  let key: CFString = "HIDIdleTime" as CFString
+  let value = CFDictionaryGetValue(dict, Unmanaged.passUnretained(key).toOpaque())
+  let number: CFNumber = unsafeBitCast(value, to: CFNumber.self)
+  var nanoseconds: Int64 = 0
+  guard CFNumberGetValue(number, CFNumberType.sInt64Type, &nanoseconds) else { return nil }
+  let interval = Double(nanoseconds) / Double(NSEC_PER_SEC)
+
+  return interval
+}
```
ErikBjare, line 285: We already have this in aw-watcher-afk, so not needed here: https://github.com/ActivityWatch/aw-watcher-afk/blob/master/aw_watcher_afk/macos.py

However, I'd be interested in any differences in behavior the implementations might have (I can't really tell).

aw_watcher_window/macos.swift
```diff
@@ -0,0 +1,285 @@
...
+    name: NSWorkspace.didActivateApplicationNotification,
+    object: nil)
+  main.focusedAppChanged()
+  detectIdle()
+}
+
+func detectIdle() {
+  let seconds = 15.0 - SystemIdleTime()!
+  if seconds < 0.0 {
+    sendHeartbeat(data: NetworkMessage(app: "", title: ""))
```
ErikBjare, line 80: Why send an empty one? Won't this cause issues?

ErikBjare (commented):
Very nice! 

Just need to strip out the AFK detection and fix some stuff surrounding how heartbeats are used.

aw_watcher_window/macos.swift
modderme123: This has two differences: 
1) it uses the iokit framework which is more modern (and more complicated) to detect time since last idle
2) It is optimized for efficiently (low cpu) getting idle time and unidle time down to the ms. This means it doesn't poll every second, instead it sleeps for the exact amount of time before idle (currently set to 15 seconds, but could be set to any number and it will only check when it needs to), and then once idle registers a callback that will detect wakeups immediately

aw_watcher_window/macos.swift
modderme123: The way I've structured this is to send an empty heartbeat as a representation of being idle because that gives precise timing of when your computer is idle instead of just when the pulseTime expires.

The swift code is designed to always send a message on change, so what happens is when you change apps it immediately sends a heartbeat to extend the last network request until now, then it sends a heartbeat to start the updated request. Because of that, I need to send something to signify that you are going idle immediately, and an empty title displays as shorter in the ui which works nicely (maybe I could send pulseTime=0 for idle instead?)

aw_watcher_window/macos.swift
modderme123: Is there a websocket based api to use the automatic ping/pong to do keep alive?

aw_watcher_window/macos.swift
ErikBjare: > maybe I could send pulseTime=0 for idle instead?

Yes, if I understand you correctly I think this would accomplish the same thing.

Also, if you detect the idle state after 15s, you should send a idle-heartbeat starting 15s ago (or whenever activity was last detected), not after the timeout has passed.

ErikBjare:
Got my M2 Air last week, so I just cleaned this up a bit, and I think it's ready for merge! Not completely ready for prod yet (although works great for me), but I've hidden it under a config flag for now.

Thank you so much for this @modderme123. Thanks to you we'll (hopefully) have a working macOS release soon! You da MVP! 🥇

ErikBjare:
@iloveitaly I've got this working now, would love for you to try it out!

ErikBjare:
Hey @modderme123. Thanks again, you're a legend for doing this. 

By modifying your changes I seem to have managed to fix all outstanding issues on macOS. Incredible, huge props!

I just released the changes in v0.12.1, enjoy!

PR #69, opened 2022-09-27
fix: misc fixes to swift script on macos, incl URL inspection

iloveitaly:
- Missing clean statement in makefile
- Extract url from safari and chrome-like browsers

The new macos window watcher is working great so far! Will do some more testing over the next couple weeks.

### DIFF
--- a/aw_watcher_window/macos.swift
+++ b/aw_watcher_window/macos.swift
@@ -110,6 +110,10 @@ func log(_ msg: String) {
   print("\(logPrefix("INFO")) \(msg)")
 }
 
+func error(_ msg: String) {
+  print("\(logPrefix("ERROR")) \(msg)")
+}
+
 // Placeholder values, set in start() from CLI arguments
 var baseurl = "http://localhost:5600"
 // NOTE: this differs from the hostname we get from Python, here we get `.local`, but in Python we get `.localdomain`
@@ -261,6 +265,14 @@ class MainThing {
   var observer: AXObserver?
   var oldWindow: AXUIElement?
 
+  // list of chrome equivalent browsers
+  let CHROME_BROWSERS = [
+    "Google Chrome",
+    "Google Chrome Canary",
+    "Chromium",
+    "Brave Browser",
+  ]
+
   func windowTitleChanged(
     _ axObserver: AXObserver,
     axElement: AXUIElement,
@@ -277,15 +289,7 @@ class MainThing {
 
     var data = NetworkMessage(app: frontmost.localizedName!, title: windowTitle as? String ?? "")
 
-    // list of chrome equivalent browsers
-    let chromeBrowsers = [
-      "Google Chrome",
-      "Google Chrome Canary",
-      "Chromium",
-      "Brave Browser",
-    ]
-
-    if chromeBrowsers.contains(frontmost.localizedName!) {
+    if CHROME_BROWSERS.contains(frontmost.localizedName!) {
       debug("Chrome browser detected, extracting URL and title")
 
       let chromeObject: ChromeProtocol = SBApplication.init(bundleIdentifier: bundleIdentifier)!
@@ -297,7 +301,14 @@ class MainThing {
         data = NetworkMessage(app: "", title: "")
       } else {
         data.url = activeTab.URL
-        if let title = activeTab.title { data.title = title }
+
+        // the tab title is more accurate and often different than the window title
+        // however, in some cases the binary does not have the right permissions to read
+        // the title properly and will return a blank string
+
+        if let tabTitle = activeTab.title {
+          if(tabTitle != "") { data.title = tabTitle }
+        }
       }
     } else if frontmost.localizedName == "Safari" {
       debug("Safari browser detected, extracting URL and title")
@@ -309,7 +320,12 @@ class MainThing {
 
       // Safari doesn't allow incognito mode to be inspected, so we do not know if we should hide the url
       data.url = activeTab.URL
-      if let title = activeTab.name { data.title = title }
+
+      // comment above applies here as well
+      if let tabTitle = activeTab.name {
+        if tabTitle != "" { data.title = tabTitle }
+      }
+
     }
 
     let heartbeat = Heartbeat(timestamp: nowTime, data: data)
--- a/aw_watcher_window/macos.swift
+++ b/aw_watcher_window/macos.swift
@@ -214,6 +214,7 @@ func sendHeartbeat(_ heartbeat: Heartbeat) {
 
         let refreshedOldHeartbeat = Heartbeat(
           // it is important to refresh the hearbeat using the timestamp where the user stopped working on the previous application
+          // more info: https://github.com/ActivityWatch/aw-watcher-window/pull/69
           // TODO unsure if this is necessary: does the timestamp of the old vs new event need to be distinct
           timestamp: heartbeat.timestamp - 1,
           data: oldHeartbeat!.data
--- a/aw_watcher_window/macos.swift
+++ b/aw_watcher_window/macos.swift
@@ -196,7 +196,7 @@ func sendHeartbeat(_ heartbeat: Heartbeat) {
 
   // if you resize a window a ton of events (subsecond) will be fired
   // we enforce a 1s minimum gap between events to avoid this
-  if timeSinceLastHeartbeat != -1.0 && timeSinceLastHeartbeat <= 1.0 {
+  if timeSinceLastHeartbeat != -1.0 && timeSinceLastHeartbeat <= 0.5 {
     debug("skipping heartbeat, last heartbeat was sent 1s ago")
     return
   }
--- a/aw_watcher_window/macos.swift
+++ b/aw_watcher_window/macos.swift
@@ -45,23 +45,9 @@ extension SBApplication: ChromeProtocol {}
     @objc optional var name: String { get } // The title of the window.
     @objc optional func id() -> Int // The unique identifier of the window.
     @objc optional var index: Int { get } // The index of the window, ordered front to back.
-    @objc optional var bounds: NSRect { get } // The bounding rectangle of the window.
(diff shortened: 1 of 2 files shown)


iloveitaly:
@ErikBjare @modderme123  mind taking a look at this one?

modderme123:
Looks good, although if you are removing the idle detection loop you might be able to remove the system idle time function as well

iloveitaly:
Nice tip! Just removed in the latest commit.

@modderme123 do you recall what this comment is referring to?

>TODO: set proper pulsetime

What's not right about the current pulsetime setup?

modderme123:
Here's how I understand pulsetime is used: https://docs.activitywatch.net/en/latest/_modules/aw_transform/heartbeats.html#heartbeat_reduce

This is very much intended for a polling system so that it doesn't ever need to diff anything, and can just send the latest value, and it will get merged if the pulsetime is long enough.

Because we intentionally don't poll, pulsetime doesn't make as much sense for us. On a new event, we just want to make sure the previous event duration is extended until the start of our new event, and the new event is created ASAP.

To extend the previous duration, we send the previous event again (except at the current time), with a pulsetime long enough that it gets merged. Then it doesn't really matter what the new event pulsetime is, because it shouldn't get merged (it necessarily has a different value, because our notifier was triggered).

I assume the original comment added by @ErikBjare was referring to wanting to switch partially over to a hyrbrid polling model like the chrome extension does (use a notifier for precise updates and then send them on a timeout, which seems like it accidentally loses resolution if you rapidly switch between apps)...

aw_watcher_window/macos.swift
```diff
@@ -85,13 +72,45 @@ struct Heartbeat: Codable {
...
   var hostname: String
 }
 
-// Placeholder values, set in start()
+// there's no builtin logging library on macos which has levels & hits stdout, so we build our own simple one
+// there a complex open source one, but it makes it harder to compile this simple one-file swift application
+
+func logTimestamp() -> String {
+  let now = Date()
+  let formatter = DateFormatter()
```
modderme123, line 90: Maybe extract this out so that you don't create a formatter every time you want to format a date?

iloveitaly:
@modderme123 thanks for the comment, very helpful! 

@ErikBjare this is ready for review and is working great for me locally. The main changes here are:

* improved the logging with levels & timestamp
* extract url information from safari
* refactored logic and added lots more comments. I _think_ I broke something when I refactored, which is what caused the conversation in discord.

aw_watcher_window/macos.swift
```diff
@@ -225,13 +298,26 @@ class MainThing {
...
       }
+    } else if frontmost.localizedName == "Safari" {
+      debug("Safari browser detected, extracting URL and title")
+
+      let safariObject: SafariApplication = SBApplication.init(bundleIdentifier: bundleIdentifier)!
+
+      let frontWindow = safariObject.windows!()[0]
+      let activeTab = frontWindow.currentTab!
+
+      // Safari doesn't have an incognito mode, we cannot hide the url
```
modderme123, line 309: This isn't true, we just need to figure out how to tell if the current window is in private browsing mode

modderme123: For example, using menus: https://alexwlchan.net/2021/04/detect-private-browsing/

iloveitaly: Sorry, comment was bad. Here's what I meant to say, which I _think_ is true:

```suggestion
      // Safari doesn't allow incognito mode to be inspected, so we do not know if we should hide the url
```

aw_watcher_window/macos.swift
```diff
@@ -158,54 +184,60 @@ func createBucket() {
...
-    } else {
-      return Optional<Heartbeat>.none
-    }
-  }
+  let oldPayloadDifferent = oldHeartbeat != nil && oldHeartbeat!.data != heartbeat.data
+  let timeSinceLastHeartbeat = oldHeartbeat != nil ? heartbeat.timestamp.timeIntervalSince(oldHeartbeat!.timestamp) : -1.0
 
-  // Then, send heartbeat starting new event
-  let payload_new = heartbeat
+  // if you resize a window a ton of events (subsecond) will be fired
```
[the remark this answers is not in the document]
iloveitaly: Ah, this is definitely better. We can compare app & title instead—I'll do this.

aw_watcher_window/macos.swift
```diff
@@ -158,54 +184,60 @@ func createBucket() {
...
       do {
-        try await sendHeartbeatSingle(payload_old, pulsetime: since_last_seconds);
+        // unlike the python aw-client library, we do not enforce a `commit_interval` and instead send the old event (which is not invalid)
+        // at the current time with a pulse value equal to the time since this event was originally sent. The aw-server will then merge
+        // this new event with the original event, extending the recorded time spent on this particular window/application.
+
+        let refreshedOldHeartbeat = Heartbeat(
+          // it is important to refresh the hearbeat using the timestamp where the user stopped working on the previous application
+          // TODO unsure if this is necessary: does the timestamp of the old vs new event need to be distinct
+          timestamp: heartbeat.timestamp - 1,
```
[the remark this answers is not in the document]
iloveitaly: What I wasn't certain about here if we can send two events with the same timestamp (i.e. the old, refreshed heartbeat) and the new heartbeat. Sounds like you didn't have any issues with that?

(this was me being lazy and not testng)

aw_watcher_window/macos.swift
iloveitaly: Good call, I'll make that change.

aw_watcher_window/macos.swift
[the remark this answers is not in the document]
modderme123: Oh I understand what this is doing now. Are the units ms or s? This seems reasonable if it is only ms

ErikBjare:
@iloveitaly Waiting for you to address @modderme123's comments :)

aw_watcher_window/macos.swift
```diff
@@ -342,6 +390,7 @@ class MainThing {
   }
 }
 
+// TODO I believe this is handled by the python wrapper so it isn't needed here
```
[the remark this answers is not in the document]
iloveitaly: Yeah, I don't think it's a massive deal either way just additional overhead to make sure they are hitting the right server APIs.

aw_watcher_window/macos.swift
iloveitaly: good call, added an issue for this here https://github.com/ActivityWatch/aw-watcher-window/issues/71

iloveitaly:
@ErikBjare @modderme123 comments addressed!

aw_watcher_window/macos.swift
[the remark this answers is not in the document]
ErikBjare: >Do we know if we can send two events with the same timestamp without causing any issues?

It *shouldn't* cause issues, but I've guarded for similar situations in other places by subtracting/adding 1ms to events.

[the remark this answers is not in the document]
iloveitaly: Got it, I'll drop it to a millisecond subtraction and call it a day. I think all comments are addressed—good to merge!

ErikBjare:
@iloveitaly Before merging this, does this resolve https://github.com/ActivityWatch/activitywatch/issues/787?

iloveitaly:
@ErikBjare I _think_ so, although I wasn't running into the identical issue he experienced. My window bucket was empty, but I had no issues with accessibility permissions. I've been running this for a couple weeks at this point and my window activity is back to normal on macos.

ErikBjare:
Alright, I tested it and it worked for both Chrome and Safari.

However, I decided to dig a bit deeper, and removed the Automation permissions for my terminal (Alacritty). After that, titles and URLs were blank again.

At minimum, it feels like we should at least get a window title even without Automation permissions, right?

Screenshot 2022-10-26 at 16 42 17

(took the screenshot after I enabled again, at which point it started working again)

aw_watcher_window/macos.swift
```diff
@@ -225,13 +299,26 @@ class MainThing {
         data.url = activeTab.URL
         if let title = activeTab.title { data.title = title }
```
ErikBjare, line 300: This check needs to not only `if let title = activeTab.title { ... }`, but also if the string is empty, to prevent that bad permissions to lead to empty titles (I assume).

Is there a reason to do title assign here at all? Won't it just be the same?

aw_watcher_window/macos.swift
```diff
@@ -225,13 +299,26 @@ class MainThing {
...
+      debug("Safari browser detected, extracting URL and title")
+
+      let safariObject: SafariApplication = SBApplication.init(bundleIdentifier: bundleIdentifier)!
+
+      let frontWindow = safariObject.windows!()[0]
+      let activeTab = frontWindow.currentTab!
+
+      // Safari doesn't allow incognito mode to be inspected, so we do not know if we should hide the url
+      data.url = activeTab.URL
+      if let title = activeTab.name { data.title = title }
```
ErikBjare, line 312: Same as above comment.

ErikBjare (commented):
Trying to get to the bottom of this browser issue.

aw_watcher_window/macos.swift
iloveitaly: Not sure if there is a reason we don't just use the original title. Make a change locally to log if there is a diff and will report back.

iloveitaly: The title is different. Reason is the tab title is different than the window title in many cases. I've updated the code to conditionally use the tab title if it is not blank. Running this locally for a week and then will report back.

ErikBjare: Can you provide a short example of a differing title? Curious what kind of differences there are

ErikBjare:
Not at my Macbook, so can't test it myself, but going to eagerly assume this is good to merge.

Thanks again for the contributions @iloveitaly! Sorry for the slow merge :)

PR #83, opened 2022-12-26
perf: use win32 API instead of WMI to query process information

chengyuhui:
Update the Windows implementation to use native Windows APIs instead of WMI, which does not consume much CPU (barely noticeable when polling with an interval of 1 second).

This should also remove the dependency on `wmi` package, but I am having difficulty using Poetry, so that change is not made.

### DIFF
--- a/aw_watcher_window/windows.py
+++ b/aw_watcher_window/windows.py
@@ -1,32 +1,34 @@
 from typing import Optional
 
-import wmi
+import os
+import time
+
 import win32gui
+import win32api
 import win32process
-
-c = wmi.WMI()
-
-"""
-Much of this derived from: http://stackoverflow.com/a/14973422/965332
-"""
 
 def get_app_path(hwnd) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


ErikBjare:
This is awesome. Great work, thanks!

ErikBjare:
@chengyuhui This created a regression for applications run as administrator, which you can read about in https://github.com/ActivityWatch/activitywatch/issues/873

Are you interested in working on a fix? Namely to use the newer method, but in the cases that it fails, fallback to using wmi.

I've asked @BelKed about it [here](https://github.com/ActivityWatch/activitywatch/issues/873#issuecomment-1785894619), thought I'd ask you too :)

chengyuhui:
I no longer use AW on my Windows system since I have migrated my main desktop to Arch. You guys can go ahead and fix this :)

ErikBjare:
Congratulations on Arch-maxing! :star_struck: 

I guess I'll make a PR myself and find someone to test it later :)

PR #86, opened 2023-02-02
[macos] guards against missing notification data

iloveitaly:
this caused some crashes in my case


ErikBjare:
LGTM! Thanks again :)

Issue #85, opened 2023-01-02
[macos] `loginwindow` reported with latest macOS update


iloveitaly:
I'm running into an issue where `login window` is reporting > 1hr of usage over a week period. I think this issue is connected to AFK and how it works. 

<img width="650" alt="image" src="https://user-images.githubusercontent.com/150855/210283599-891dd9ea-40df-4773-b2d0-82d23fe71d9b.png">

1. Does `Exclude AFK time` filter out time where there is no AFK data? i.e. will the login window time above be excluded?
2. If not, do you think we should just explicitly filter out loginwindow events (string matching on the title)? Or should we improve AFK to 'fill in' the missing time period when the computer was sleeping (similar to what the macOS window watcher implementation does).

iloveitaly:
@ErikBjare friendly ping here—wdyt is the best approach?

ErikBjare:
Exclude AFK time only selects time where there is a 'not-afk' event. If data is missing or 'afk', it won't/shouldn't pass the filter.

This could possibly be some other bug with overlapping events, not sure why/how this happens.

ErikBjare:
One possibility is that on unlock, the user might keep using whatever app is active (no app/title switch to trigger a new event), and the watcher continues to think loginwindow is active.

I've had a similar issue where, for some applications, the window-switch doesn't seem to fire an event. Like when I start games from Steam, it will sometimes keep tracking Steam until I tab-switch back and forth. Adding polling to augment the event-listener would fix this. Someone else has reported the same in issues or on the Discord.

iloveitaly:
> Adding polling to augment the event-listener would fix this.

Ah, interesting. I think this is a good approach—do a poll every 10s or so if we haven't received a OS notification. Does that sound good to you?

iloveitaly:
@ErikBjare I've just hardcoded excluding loginwindow events for now and it's working pretty well. I think adding a poll to the swift implementation will be better, but I'm wondering if you are open to a hardcoded exclusion in the meantime (I don't have time for the polling stuff right now).

PR #88, opened 2023-03-07
fix: ignore loginwindow events

iloveitaly:
workaround for #85. Yes, we should do the polling solution we outlined, but this has worked for me locally the last couple weeks and I don't have the time for anything else right now :)

could we get this in as a stopgap until someone has time for the correct solution?


iloveitaly:
@ErikBjare friendly reminder here!

iloveitaly:
@ErikBjare friendly reminder here!

ErikBjare:
I'm not sure how I feel about this.

Kinda feels like sweeping the problem under the rug. Merging this PR would lead to otherwise unexplained gaps (unless you happen to look at logs).

I would prefer a workaround on the front-end, such as adding a "ignored pattern" feature (similar to the "always count as active pattern" feature). Which is a wanted feature anyway.

iloveitaly:
@ErikBjare what do you think about "filling in" AFK events as outlined in #85? It feels like we should have a AFK timeline, even if the computer goes to sleep.

This would fix this issue without (a) adding the complexity of polling in the swift app and (b) without hacky exclusion rules.

iloveitaly:
@ErikBjare friendly bump here! Curious your thoughts on the above Q.

theolundqvist:
> I'm not sure how I feel about this.
> 
> Kinda feels like sweeping the problem under the rug. Merging this PR would lead to otherwise unexplained gaps (unless you happen to look at logs).
> 
> I would prefer a workaround on the front-end, such as adding a "ignored pattern" feature (similar to the "always count as active pattern" feature). Which is a wanted feature anyway.

Hello! Is there any such feature? I would like to ignore the terminal window completely since I use the tmux watcher. Otherwise i get double the time each day. How is this usually solved when using editor watchers?

ErikBjare:
@theolundqvist It's fairly trivial to implement (esp if you look at the audible-as-active PR for guidance). You could probably do it yourself! If not up your alley, I will try to get around to it Soon™.

Also, cool to see someone in the same city as me randomly on the internet. If you wanna grab lunch sometime, let me know :)

ErikBjare:
@iloveitaly Filling in AFK events when the device is asleep/off is not something we want to do (we used to!), as it makes it very difficult to know:

 - if the computer was on
 - if the other watchers have crashed/missing data

Looking at this, I think that there might be an issue with the window watcher. If the computer is off (unable to pull the active window), it should not fill the time when the device was off.

The window watcher should not use a high `pulsetime` to fill these gaps, but instead rely on the poll-augmentation to expect a recent last event, which is updated with a smaller `pulsetime`.

PR #95, opened 2023-12-23
feat: added hybrid polling method for macOS

ErikBjare:
This implements the hybrid polling mechanism that augments the events-based method with a 10s poll, as we've discussed before (https://github.com/ActivityWatch/aw-watcher-window/pull/69#issuecomment-1264419382). 

This fixes bugs where if the starting/ending event is missed, and would've created a long continuous event, that period of time would be lost.

I generated this with [gptme](https://github.com/ErikBjare/gptme) running GPT-4.

~~No idea if it works or not, but doesn't crash anyway.~~

It didn't work at first, but it works great now! (after 1min of testing)

### DIFF
--- a/aw_watcher_window/macos.swift
+++ b/aw_watcher_window/macos.swift
@@ -177,6 +177,9 @@ func start() {
   )
 
   main.focusedAppChanged()
+
+  // Start the polling timer
+  main.pollingTimer = Timer.scheduledTimer(timeInterval: 10.0, target: main, selector: #selector(main.pollActiveWindow), userInfo: nil, repeats: true)
 }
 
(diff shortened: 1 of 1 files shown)


ErikBjare:
@iloveitaly @modderme123 This seems to fix lots of issues I've had with gaps in the data.

Please give it a review, I want to merge and release a new beta with it asap :)

ErikBjare:
After a few days of testing, it seems solid, and fixes quite a few issues with the previous implementation.

@iloveitaly @modderme123 Feel free to still give it a review post-merge :)

iloveitaly:
NICE! Glad this is finally getting fixed. Thank you!

PR #97, opened 2024-02-04
Added exception handling for `Xlib.error.BadWindow`

BasileusErwin:
Implemented a try/except block to catch the `Xlib.error.BadWindow` error during window name and class fetching. This change prevents the appearance of unnecessary messages in the logs, which although they do not affect the performance or operation of the watcher, they generate noise and do not provide relevant information about the error. With this enhancement, in case this error is detected, the window name or class will be set to "unknown", allowing the process to continue without interruption


ErikBjare:
Nice, thanks!

PR #99, opened 2024-02-04
[Feature] Add `exclude_titles` feature for enhanced window title exclusion

BasileusErwin:
This commit introduces the `--exclude-titles` argument to the aw-watcher-window module, allowing users to specify a list of window titles or regular expression patterns to exclude from tracking. This new feature is designed to complement the existing `--exclude-title` flag, providing enhanced flexibility for users who need to exclude multiple window titles without breaking compatibility with existing configurations.

Key Changes:
- Added the `--exclude-titles` argument to the argparse configuration in `config.py`, enabling the specification of multiple exclusion patterns.
- Updated the `heartbeat_loop` function in `main.py` to support both `exclude_title` and `exclude_titles`, with `exclude_titles` allowing for an array of titles to be excluded.
- Utilized the `re` module for regex pattern matching against window titles, ensuring case-insensitive comparisons.

This enhancement ensures that users can now more precisely control which window titles are excluded from tracking, making the aw-watcher-window module more versatile and user-friendly.

### DIFF
--- a/aw_watcher_window/config.py
+++ b/aw_watcher_window/config.py
@@ -40,7 +40,7 @@ def parse_args():
         dest="exclude_titles",
         nargs='+',
         default=default_exclude_titles,
-        help="List of window titles or regular expression patterns to exclude from tracking. Can specify multiple titles."
+        help="List of regular expression patterns to exclude from tracking. Each pattern can match any part of a window title. Use standard regex syntax. For example, to exclude windows containing 'Game', simply use 'Game'. Can specify multiple patterns."
     )
     parser.add_argument("--verbose", dest="verbose", action="store_true")
     parser.add_argument(
--- a/aw_watcher_window/config.py
+++ b/aw_watcher_window/config.py
@@ -40,7 +40,7 @@ def parse_args():
         dest="exclude_titles",
         nargs='+',
         default=default_exclude_titles,
-        help="List of regular expression patterns to exclude from tracking. Each pattern can match any part of a window title. Use standard regex syntax. For example, to exclude windows containing 'Game', simply use 'Game'. Can specify multiple patterns."
+        help="List of window titles or regular expression patterns to exclude from tracking. Can specify multiple titles."
     )
     parser.add_argument("--verbose", dest="verbose", action="store_true")
     parser.add_argument(
--- a/aw_watcher_window/config.py
+++ b/aw_watcher_window/config.py
@@ -19,6 +19,7 @@ def parse_args():
 
     default_poll_time = config["poll_time"]
     default_exclude_title = config["exclude_title"]
+    default_exclude_titles = config["exclude_titles"]
     default_strategy_macos = config["strategy_macos"]
 
     parser = argparse.ArgumentParser(
@@ -32,6 +33,13 @@ def parse_args():
         dest="exclude_title",
         action="store_true",
         default=default_exclude_title,
+    )
+    parser.add_argument(
+        "--exclude-titles",
+        dest="exclude_titles",
+        nargs='+',
+        default=default_exclude_titles,
+        help="List of window titles or regular expression patterns to exclude from tracking. Can specify multiple titles."
     )
     parser.add_argument("--verbose", dest="verbose", action="store_true")
     parser.add_argument(
--- a/aw_watcher_window/config.py
+++ b/aw_watcher_window/config.py
@@ -40,7 +40,7 @@ def parse_args():
         dest="exclude_titles",
         nargs='+',
         default=default_exclude_titles,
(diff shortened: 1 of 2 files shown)


aw_watcher_window/main.py
```diff
@@ -128,6 +130,12 @@ def heartbeat_loop(client, bucket_id, poll_time, strategy, exclude_title=False):
         if current_window is None:
             logger.debug("Unable to fetch window, trying again on next poll")
         else:
+            if exclude_titles:
+                for title in exclude_titles:
+                    pattern = re.compile(re.escape(title), re.IGNORECASE)
+                    if pattern.search(current_window["title"]):
+                        current_window["title"] = "excluded"
```
ErikBjare, line 137:
```suggestion
            for title in exclude_titles:
                pattern = re.compile(re.escape(title), re.IGNORECASE)
                if pattern.search(current_window["title"]):
                    current_window["title"] = "excluded"
```

aw_watcher_window/main.py
```diff
@@ -128,6 +130,12 @@ def heartbeat_loop(client, bucket_id, poll_time, strategy, exclude_title=False):
         if current_window is None:
             logger.debug("Unable to fetch window, trying again on next poll")
         else:
+            if exclude_titles:
+                for title in exclude_titles:
+                    pattern = re.compile(re.escape(title), re.IGNORECASE)
+                    if pattern.search(current_window["title"]):
```
ErikBjare, line 136: Seems you are treating all strings as regexes? This might confuse some users.

aw_watcher_window/main.py
```diff
@@ -128,6 +130,12 @@ def heartbeat_loop(client, bucket_id, poll_time, strategy, exclude_title=False):
         if current_window is None:
             logger.debug("Unable to fetch window, trying again on next poll")
         else:
+            if exclude_titles:
+                for title in exclude_titles:
+                    pattern = re.compile(re.escape(title), re.IGNORECASE)
```
ErikBjare, line 135: The regex should be compiled once, not on every loop.

aw_watcher_window/main.py
```diff
@@ -128,6 +130,11 @@ def heartbeat_loop(client, bucket_id, poll_time, strategy, exclude_title=False):
         if current_window is None:
             logger.debug("Unable to fetch window, trying again on next poll")
         else:
+            patterns = [re.compile(re.escape(title), re.IGNORECASE) for title in exclude_titles]
```
ErikBjare, line 133: You are still compiling on every (outer) loop. You need to pass the compiled regexes to the heartbeat_loop function.

ErikBjare (changes requested):

aw_watcher_window/main.py
```diff
@@ -92,10 +93,11 @@ def main():
                 poll_time=args.poll_time,
                 strategy=args.strategy,
                 exclude_title=args.exclude_title,
+                exclude_titles=[re.compile(re.escape(title), re.IGNORECASE) for title in args.exclude_titles]
```
ErikBjare, line 96: Why are you escaping this? If you are escaping, then it's not really matching a regex, as all regex-special characters (`.[]()+*?` etc) would be escaped and wouldn't work (iirc).

BasileusErwin: I did it this way because it is the way that worked for me in all cases, but it is perfectly possible to remove it

ErikBjare: What do you mean "all cases"? This breaks regex functionality completely. Perhaps you used a bad regex?

ErikBjare:
It seems Ellipsis wanted to put the compile in a try/except, which is probably a good idea.
But, unlike what it did in its PR, it should print a warning if regex fails to compile (maybe even `exit(1)`).

aw_watcher_window/main.py
BasileusErwin: It is not so much that they are bad regex, but that by putting part of a title it excludes it, but without escaping the characters it works the same.

aw_watcher_window/config.py
```diff
@@ -33,6 +35,13 @@ def parse_args():
         action="store_true",
         default=default_exclude_title,
     )
+    parser.add_argument(
+        "--exclude-titles",
+        dest="exclude_titles",
+        nargs='+',
+        default=default_exclude_titles,
+        help="List of window titles or regular expression patterns to exclude from tracking. Can specify multiple titles."
```
ErikBjare, line 43:
```suggestion
        help="List of regular expressions, matched against window titles, to exclude from tracking. Can specify multiple times."
```

ErikBjare (approved):
Nice! Thanks :)

PR #118, opened 2026-02-27
fix(windows): handle hwnd==0 gracefully for UAC/secure desktop

### LINKED ISSUE #90: [Windows] Exception thrown while trying to get active window
[Reported](https://discord.com/channels/755040852727955476/755334543891759194/1097209893779427389) by `@Creve` on Discord.

---

**Logs:**
```
2023-04-16 18:56:19 [ERROR]: Exception thrown while trying to get active window  (aw_watcher_window.main:124)
Traceback (most recent call last):
  File "aw_watcher_window\main.py", line 106, in heartbeat_loop
  File "aw_watcher_window\lib.py", line 65, in get_current_window
  File "aw_watcher_window\lib.py", line 42, in get_current_window_windows
  File "aw_watcher_window\windows.py", line 26, in get_app_name
  File "aw_watcher_window\windows.py", line 15, in get_app_path
pywintypes.error: (87, 'OpenProcess', 'Параметр задан неверно.')
```

---

- **OS:** Windows
- **ActivityWatch:** ActivityWatch `v0.12.3b1`

TimeToBuildBob:
## Summary

- Skip window polling when `hwnd == 0` (no foreground window), which occurs during UAC prompts, lock screen, and secure desktop
- Wrap the WMI fallback in try/except so if both win32api and WMI fail for elevated processes, we degrade to "unknown" instead of error spam

## Problem

When the foreground window handle is 0 (e.g. during UAC prompts, lock screen, or secure desktop), `OpenProcess` fails with `pywintypes.error: (87, 'OpenProcess', 'The parameter is incorrect.')`. This causes continuous error logging every poll cycle.

Additionally, when the primary `get_app_name()` fails for elevated/admin processes (error 5: Access denied), the WMI fallback could also fail and propagate exceptions upward, causing more error spam.

## Fix

1. **Early return for `hwnd == 0`**: `get_active_window_handle()` returns 0 when there's no foreground window. We now return `None` early, which `heartbeat_loop` already handles gracefully (logs debug, skips heartbeat, tries again next poll).

2. **Wrapped WMI fallback**: If both `get_app_name()` and `get_app_name_wmi()` fail, we now set `app = None` which resolves to `"unknown"` instead of propagating the exception.

## Test plan

- [x] Verified `heartbeat_loop` already handles `None` return from `get_current_window()` (line 144-145 in main.py)
- [ ] Manual test on Windows with UAC prompt active
- [ ] Manual test running admin-elevated application (e.g. game with admin rights)

Fixes #90

----

> [!IMPORTANT]
> Handles `hwnd == 0` gracefully in `get_current_window_windows()` and wraps WMI fallback to prevent error spam in `lib.py`.
> 
>   - **Behavior**:
>     - `get_current_window_windows()` in `lib.py` returns `None` if `window_handle` is 0, skipping the poll cycle during UAC prompts or secure desktop.
>     - Wraps WMI fallback in `get_current_window_windows()` to set `app = None` if both `get_app_name()` and `get_app_name_wmi()` fail, resolving to "unknown" instead of propagating exceptions.
>   - **Misc**:
>     - No changes to Linux or macOS functions in `lib.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ActivityWatch%2Faw-watcher-window&utm_source=github&utm_medium=referral)<sup> for 0723a64b01cb77c6b978e24022550fddacd00709. You can [customize](https://app.ellipsis.dev/ActivityWatch/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/aw_watcher_window/lib.py
+++ b/aw_watcher_window/lib.py
@@ -39,11 +39,20 @@ def get_current_window_windows() -> Optional[dict]:
     from . import windows
 
     window_handle = windows.get_active_window_handle()
+
+    # hwnd 0 means no foreground window (e.g. during UAC prompt, lock screen,
+    # or secure desktop). Return None so heartbeat_loop skips this poll cycle.
+    if not window_handle:
+        return None
+
     try:
         app = windows.get_app_name(window_handle)
-    except Exception:  # TODO: narrow down the exception
-        # try with wmi method
-        app = windows.get_app_name_wmi(window_handle)
+    except Exception:
+        # Fall back to WMI for elevated/admin processes where OpenProcess fails
+        try:
+            app = windows.get_app_name_wmi(window_handle)
+        except Exception:
+            app = None
 
     title = windows.get_window_title(window_handle)
 


aw_watcher_window/lib.py
```diff
@@ -39,11 +39,20 @@ def get_current_window_windows() -> Optional[dict]:
     from . import windows
 
     window_handle = windows.get_active_window_handle()
+
+    # hwnd 0 means no foreground window (e.g. during UAC prompt, lock screen,
+    # or secure desktop). Return None so heartbeat_loop skips this poll cycle.
```
ErikBjare, line 44: Should it skip poll cycle or lead to unknown?

ErikBjare:
@TimeToBuildBob See https://github.com/ActivityWatch/aw-watcher-window/pull/118#discussion_r2863812999

aw_watcher_window/lib.py
TimeToBuildBob: Good question — I'd argue **skip the poll cycle** is the right behavior here, for the same reason we already do it for non-Windows (the `get_current_window` functions return `None` and `heartbeat_loop` treats it as 'no window, try again').

The distinction:
- `hwnd == 0`: There is literally **no foreground window** (lock screen, UAC prompt, secure desktop). Nothing to record.
- `app = None` (from failed OpenProcess / WMI): There **is** a window, we just can't identify it → falls back to `"unknown"`.

Recording `"unknown"` during a UAC prompt or lock screen would create misleading activity — the user isn't actually in any app. Skipping the heartbeat leaves a gap (or the previous heartbeat's pulsetime covers it), which is semantically correct.

That said, if you'd prefer `"unknown"` for continuity, happy to change it — it's a one-liner swap.

Issue #125, opened 2026-05-07
`v0.14.0b1` is built for macOS 15.0


BelKed:
Some parts of the latest beta (`v0.14.0b1`) seems to be supporting macOS Sonoma (14.8.2) but `aw-watcher-window` crashes on this version

---

```console
Module aw-watcher-window quit unexpectedly after 3 auto-restart attempts in 10 minutes

2026-05-07 15:51:19 [INFO ]: aw-watcher-window started  (aw_watcher_window.main:71)
2026-05-07 15:51:19 [INFO ]: Using swift strategy, calling out to swift binary  (aw_watcher_window.main:76)
2026-05-07 15:51:19 [INFO ]: Connection to aw-server established by aw-watcher-window  (aw_client.client:470)
```

---

```console
$ /Applications/ActivityWatch.app/Contents/MacOS/aw-watcher-window
2026-05-07 15:55:31 [INFO ]: aw-watcher-window started  (aw_watcher_window.main:71)
2026-05-07 15:55:31 [INFO ]: Using swift strategy, calling out to swift binary  (aw_watcher_window.main:76)
2026-05-07 15:55:31 [INFO ]: Connection to aw-server established by aw-watcher-window  (aw_client.client:470)
dyld[71919]: Library not loaded: /usr/lib/swift/libswift_stdio.dylib
  Referenced from: <BC5E953F-259D-3C3D-A37B-412247E96A48> /Applications/ActivityWatch.app/Contents/Frameworks/aw_watcher_window/aw-watcher-window-macos (built for macOS 15.0 which is newer than running OS)
  Reason: tried: '/usr/lib/swift/libswift_stdio.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/lib/swift/libswift_stdio.dylib' (no such file), '/usr/lib/swift/libswift_stdio.dylib' (no such file, not in dyld cache)
usage: aw-watcher-window [-h] [--host HOST] [--port PORT] [--testing] [--exclude-title] [--exclude-titles EXCLUDE_TITLES [EXCLUDE_TITLES ...]] [--verbose]
                         [--poll-time POLL_TIME] [--strategy {jxa,applescript,swift}]
aw-watcher-window: error: unrecognized arguments: -B -S -I -c from multiprocessing.resource_tracker import main;main(6)
[...]

PR #132, opened 2026-07-22
fix(macos): lower swift helper deployment target from 14.0 to 12.0

ErikBjare:
## Why

The 14.0 target from #126 was chosen to stop the runner-default target (macOS 15) from leaking into the binary — not because the code needs macOS 14 APIs. The measured floor is **macOS 12**: Swift concurrency (`Task`) needs 10.15 and `Date.now` needs 12. The helper compiles cleanly at 12.0 for both x86_64 and arm64, and fails at 11.0 and below.

Lowering to 12.0:
- keeps the helper usable on Monterey-era machines (Intel Macs back to ~2014–15), ahead of restoring x86_64 bundle builds (ActivityWatch/activitywatch#1263)
- matches the bundle repo, which now builds everything with `MACOSX_DEPLOYMENT_TARGET=12.0` (ActivityWatch/activitywatch#1363) — previously the bundle's env override (`?=`) and this repo's default/CI check disagreed about the floor
- still fixes the #125 Sonoma dyld crash, which only required an explicit target ≤ the user's OS

## Changes

- `Makefile`: `MACOSX_DEPLOYMENT_TARGET ?= 14.0` → `12.0`
- `.github/workflows/build.yml`: build/package env and the `otool` `minos` check updated to match

Verified locally: `swiftc -target arm64-apple-macosx12.0` builds and `otool -l` reports `minos 12.0`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
         pip install poetry==1.3.2  
         source venv/bin/activate || source venv/Scripts/activate
         if [ "${{ runner.os }}" = "macOS" ]; then
-          MACOSX_DEPLOYMENT_TARGET=14.0 make build
+          MACOSX_DEPLOYMENT_TARGET=12.0 make build
         else
           make build
         fi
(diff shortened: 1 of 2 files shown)


TimeToBuildBob (approved):
Reviewed the full diff and CI. This correctly makes 12.0 the single default across the Makefile, macOS build/package jobs, and the Mach-O `minos` assertion. The measured API floor and ActivityWatch/activitywatch#1363 now agree; all CI is green and Greptile is 5/5. LGTM.

ErikBjare:
@greptileai review

PR #135, opened 2026-07-31
feat(privacy): add watcher-side privacy filter (drop/redact before send)

TimeToBuildBob:
## Summary

Adds a client-side privacy filter that drops or redacts window events **before they are sent to aw-server**, so sensitive data never leaves the machine at all.

This mirrors the server-side `privacy_filters` engine that landed in aw-server-rust#600, allowing users to enforce the same rules at both layers. Previously the watcher only supported dropping titles via `exclude_title`/`exclude_titles` (redact to `"excluded"`). This PR adds explicit `drop` and `redact` actions with user-configurable regex patterns and replacement strings.

## What changed

- **`aw_watcher_window/privacy_filter.py`** (new) — pure Python, no new deps:
  - `compile_privacy_rules(raw)` — validates and compiles regex patterns; invalid regexes and unknown actions are skipped with an error log rather than crashing
  - `apply_privacy_filters(window, rules)` — returns `None` (drop) or a filtered copy (redact); never mutates the input dict
- **`aw_watcher_window/config.py`** — exposes `privacy_filter_rules` from `[[aw-watcher-window.privacy_filter]]` TOML array of tables
- **`aw_watcher_window/main.py`** — privacy filter runs first in `transform_window`; a `None` return skips the heartbeat; `heartbeat_loop` threads the rules through
- **`tests/test_privacy_filter.py`** (new) — 23 tests covering compile, drop, redact, input immutability, edge cases
- **`README.md`** — documents the new config section with examples

## Config example

```toml
# Drop private-browsing windows entirely
[[aw-watcher-window.privacy_filter]]
pattern = "(?i)private browsing|incognito"
action  = "drop"

# Redact banking titles
[[aw-watcher-window.privacy_filter]]
pattern     = "(?i)bank|my account"
action      = "redact"
replacement = "REDACTED"   # optional; defaults to "excluded"

# Drop by app name (not title)
[[aw-watcher-window.privacy_filter]]
pattern = "(?i)signal"
field   = "app"
action  = "drop"
```

Rule fields: `pattern` (Python regex), `field` (default `"title"`), `action` (`"drop"` or `"redact"`), `replacement` (for redact; default `"excluded"`).

## Ordering

Privacy filter runs **before** Research Edition and `exclude_title` transforms. A `"drop"` rule exits immediately — subsequent rules are not evaluated for that event.

## macOS note

The default `swift` strategy bypasses this Python transform (same as Research Edition). Use `--strategy jxa` or `--strategy applescript` to enable watcher-side privacy filtering on macOS.

## Test plan

- [x] `python3 -m pytest tests/test_privacy_filter.py -v` — 23 passed
- [x] `python3 -m pytest tests/test_main.py -v` — 12 passed (no regressions)
- [x] Normal run with no config: zero behaviour change
- [x] `drop` rule matching title: heartbeat skipped
- [x] `redact` rule matching title: heartbeat sent with replacement value
- [x] Invalid regex in rule: rule skipped with error log, watcher continues

### DIFF
--- a/aw_watcher_window/main.py
+++ b/aw_watcher_window/main.py
@@ -14,6 +14,7 @@ from aw_core.models import Event
 from .config import parse_args
 from .exceptions import FatalError
 from .lib import get_current_window
+from .privacy_filter import apply_privacy_filters, compile_privacy_rules
 from .research_filter import transform as research_transform
 from .macos_cli import build_swift_command
 from .macos_permissions import background_ensure_permissions
@@ -95,6 +96,9 @@ def main():
             if args.research_enabled
             else None
         )
+        privacy_filter_rules = compile_privacy_rules(
+            getattr(args, "privacy_filter_rules", [])
+        )
         if sys.platform == "darwin" and args.strategy == "swift":
             logger.info("Using swift strategy, calling out to swift binary")
             binpath = os.path.join(
@@ -135,6 +139,7 @@ def main():
                 ],
                 research_category_map=research_category_map,
                 research_app_category_map=research_app_category_map,
+                privacy_filter_rules=privacy_filter_rules,
             )
 
 
@@ -147,6 +152,7 @@ def heartbeat_loop(
     exclude_titles=[],
     research_category_map=None,
     research_app_category_map=None,
+    privacy_filter_rules=None,
 ):
     while True:
         if os.getppid() == 1:
@@ -186,17 +192,21 @@ def heartbeat_loop(
                 exclude_titles=exclude_titles,
                 research_category_map=research_category_map,
                 research_app_category_map=research_app_category_map,
+                privacy_filter_rules=privacy_filter_rules,
             )
 
-            now = datetime.now(timezone.utc)
-            current_window_event = Event(timestamp=now, data=current_window)
+            if current_window is None:
+                logger.debug("Event dropped by privacy filter, skipping heartbeat")
+            else:
+                now = datetime.now(timezone.utc)
+                current_window_event = Event(timestamp=now, data=current_window)
 
-            client.heartbeat(
-                bucket_id,
-                current_window_event,
-                pulsetime=compute_pulsetime(poll_time),
-                queued=True,
-            )
+                client.heartbeat(
+                    bucket_id,
+                    current_window_event,
+                    pulsetime=compute_pulsetime(poll_time),
+                    queued=True,
+                )
 
         sleep(poll_time)
 
@@ -207,7 +217,14 @@ def transform_window(
     exclude_titles=None,
     research_category_map=None,
     research_app_category_map=None,
+    privacy_filter_rules=None,
 ):
+    # Privacy filter runs first — sensitive events never reach other transforms
+    if privacy_filter_rules:
+        current_window = apply_privacy_filters(current_window, privacy_filter_rules)
+        if current_window is None:
+            return None
+
     if research_category_map is not None:
         return research_transform(
             current_window,
--- a/aw_watcher_window/privacy_filter.py
+++ b/aw_watcher_window/privacy_filter.py
@@ -58,7 +58,7 @@ def compile_privacy_rules(raw_rules: list) -> list:
 
         try:
             pattern = re.compile(pattern_str)
-        except re.error as exc:
+        except (re.error, TypeError) as exc:
             logger.error(
                 "privacy_filter: invalid regex %r — %s — rule skipped", pattern_str, exc
             )
--- a/aw_watcher_window/privacy_filter.py
+++ b/aw_watcher_window/privacy_filter.py
@@ -1,0 +1,116 @@
+"""
+Watcher-side privacy filter: drop or redact sensitive events before sending.
+
+Mirrors the server-side ``privacy_filters`` engine in aw-server-rust so that
+sensitive data never leaves the machine in the first place.
+
+Configure rules in ``~/.config/activitywatch/aw-watcher-window/aw-watcher-window.toml``:
+
+.. code-block:: toml
+
+   # Drop events whose title matches a private-browsing pattern
+   [[aw-watcher-window.privacy_filter]]
+   pattern = "(?i)private browsing|incognito"
+   action  = "drop"
+
+   # Redact window titles containing a banking domain
+   [[aw-watcher-window.privacy_filter]]
+   pattern     = "(?i)bank\\.example\\.com"
+   action      = "redact"
+   replacement = "REDACTED"   # optional; defaults to "excluded"
+
+Rule fields:
+  pattern     -- Python ``re`` regex applied to the target field value.
+  field       -- Window-event field to match against (default: ``"title"``).
+  action      -- ``"drop"`` (discard the event) or ``"redact"`` (replace the value).
+  replacement -- Replacement string for ``"redact"`` action (default: ``"excluded"``).
+
+macOS note: the default ``swift`` strategy bypasses this Python transform.
+Use ``--strategy jxa`` or ``--strategy applescript`` if watcher-side privacy
+filtering is required on macOS.
+"""
+
+import logging
+import re
+from typing import Optional
+
+logger = logging.getLogger(__name__)
+
+_VALID_ACTIONS = frozenset({"drop", "redact"})
+
+
+def compile_privacy_rules(raw_rules: list) -> list:
+    """Validate and compile regex patterns in privacy filter rules.
+
+    Returns a list of compiled rule dicts ready for :func:`apply_privacy_filters`.
+    Invalid rules (bad regex or unknown action) are skipped with an error log.
+    """
+    compiled = []
+    for raw in raw_rules:
+        if not isinstance(raw, dict):
+            logger.error("privacy_filter rule is not a table: %r — skipped", raw)
+            continue
+
+        pattern_str = raw.get("pattern", "")
+        if not pattern_str:
+            logger.error("privacy_filter rule missing 'pattern' — skipped: %r", raw)
+            continue
+
+        try:
+            pattern = re.compile(pattern_str)
+        except re.error as exc:
+            logger.error(
+                "privacy_filter: invalid regex %r — %s — rule skipped", pattern_str, exc
+            )
+            continue
+
+        action = raw.get("action", "redact")
+        if action not in _VALID_ACTIONS:
+            logger.error(
+                "privacy_filter: unknown action %r (expected 'drop' or 'redact') — rule skipped",
+                action,
+            )
+            continue
+
+        compiled.append(
+            {
+                "pattern": pattern,
+                "field": raw.get("field", "title"),
+                "action": action,
+                "replacement": raw.get("replacement", "excluded"),
+            }
+        )
+    return compiled
+
+
+def apply_privacy_filters(window: dict, rules: list) -> Optional[dict]:
+    """Apply compiled privacy filter rules to a window event.
+
+    Returns ``None`` when the event should be dropped entirely, otherwise a
+    (possibly modified) copy of the window dict. Never mutates the input.
+    """
+    if not rules:
+        return window
+
+    result = dict(window)
+    for rule in rules:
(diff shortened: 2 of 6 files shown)


aw_watcher_window/privacy_filter.py
```diff
@@ -0,0 +1,116 @@
...
+            continue
+
+        pattern_str = raw.get("pattern", "")
+        if not pattern_str:
+            logger.error("privacy_filter rule missing 'pattern' — skipped: %r", raw)
+            continue
+
+        try:
+            pattern = re.compile(pattern_str)
+        except re.error as exc:
```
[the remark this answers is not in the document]
TimeToBuildBob: Good catch. Fixed in [`cdf5c3b`](https://github.com/ActivityWatch/aw-watcher-window/commit/cdf5c3b4024800b3f0609a28e0d2c2f065f98561): `compile_privacy_rules` now catches `TypeError` from non-string TOML pattern values and skips the malformed rule. Added a regression test for `pattern = 123`; all 24 focused privacy-filter tests pass.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob The watcher-side privacy filter should also be configurable via webui (non-technical user-facing central config) similar to the server-side privacy filter (but watcher-only config should also be supported, if e.g. server is untrusted).

TimeToBuildBob:
Good point. The TOML config in this PR covers the watcher-only path you described (useful when the server is untrusted or the user wants filtering before data ever leaves the machine).

For webui configurability, the shape I'd expect is:
- aw-server exposes a config endpoint for watcher-side privacy rules (similar to how server-side `privacy_filters` are managed)
- aw-webui adds a settings panel for configuring those rules
- The watcher fetches its config from the server when available, falling back to TOML when disconnected or server is untrusted

That touches aw-server and aw-webui (separate repos), so it's naturally a follow-up rather than something that fits cleanly into this PR. Would you prefer I scope it here (which would expand the PR significantly) or open a follow-up issue to track it?

Also resolved the rebase conflict with master (#136, #137) — the branch is up to date now.

TimeToBuildBob:
Still waiting for direction on scope — whether to include the webui-configurable path in this PR or track it as a follow-up. The watcher-only TOML config path is ready, CI green, and rebased. Happy to expand or split depending on your preference.

TimeToBuildBob:
CI-green and mergeable (Greptile 5/5) — **waiting only on a maintainer click**.

This PR is ready to merge, but the bot has pull-only access to this repo and can't self-merge — surfacing it here so it isn't lost. The monitoring loop will stop re-flagging it now that this note is posted.

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 2 findings disposed (fixed in 1d68a6b; rejected)

<sub>Updated after inline dispositions on finding threads — this is the current state; the verdict below is frozen at review time and is kept as the historical record of that pass.</sub>

<details><summary>Finding disposition</summary>

| Finding | Severity | State |
| --- | --- | --- |
| `aw_watcher_window/main.py:102` | P1 | fixed in `1d68a6b` |
| `aw_watcher_window/main.py:224` | P1 | rejected — The verification said: "the exclude_titles loop at main.py line 235-237 mutates the original current_window, but the fun |
| `aw_watcher_window/main.py:102` | P2 | rejected |

</details>




Adds a watcher-side privacy filter that drops or redacts window events before they are sent to aw-server. New privacy_filter.py compiles and applies regex rules; config.py reads [[aw-watcher-window.privacy_filter]] TOML tables; main.py runs the filter first in transform_window and skips heartbeats for dropped events, and refuses to start under the macOS swift strategy when rules are configured. Adds tests for the filter, config parsing, and the swift refusal, plus README documentation and CodeQL action version bumps.

### Needs a look — P2 only

Confidence 4/5

<details><summary><b>1 finding</b> · ⚠️ 1 P2</summary>

**⚠️ P2 medium** — `aw_watcher_window/main.py:102`

The swift strategy refusal check only runs when sys.platform == 'darwin' and args.strategy == 'swift', but the default strategy on macOS is 'swift' (config default strategy_macos = "swift"). However, the check is placed after compile_privacy_rules, which is fine. The real issue: the refusal only happens in main() when the Python process starts. But the swift strategy is also the default for macOS, and the README says the watcher refuses to start. The check is correct. Wait, let me look for actual bugs. The privacy filter runs in transform_window, which is only called in the Python heartbeat_loop path. The swift strategy path calls build_swift_command and Popen, and the refusal happens before that. That's fine. But what about the case where privacy_filter_rules is configured but the swift strategy is used on a non-macOS platform? The check is guarded by sys.platform == 'darwin

<sub>**How this was verified:** Checked main() flow: compile_privacy_rules at line 99, then the swift check at line 102. On non-darwin, the else branch calls heartbeat_loop, which applies the filter. On darwin with swift, it exits. No bug here.</sub>

</details>

<details><summary><b>Files changed (8)</b> — the diff as I read it</summary>

- `.github/workflows/codeql.yml` — Bumps actions/checkout from v3 to v4 and codeql-action from v2 to v3.
- `README.md` — Documents the new privacy_filter config section with examples and a macOS swift strategy note.
- `aw_watcher_window/config.py` — Reads privacy_filter from config, wrapping a single TOML table into a list, and attaches privacy_filter_rules to parsed args.
- `aw_watcher_window/main.py` — Compiles privacy rules, refuses to start under swift strategy when rules exist, threads rules through heartbeat_loop, and skips heartbeat when transform returns None.
- `aw_watcher_window/privacy_filter.py` — New module with compile_privacy_rules and apply_privacy_filters for drop/redact actions.
- `tests/test_config.py` — Adds tests for wrapping a single privacy_filter table and preserving an array of rules.
- `tests/test_main.py` — Adds tests for exclude_titles interaction with privacy rules and swift refusal to start.
- `tests/test_privacy_filter.py` — New test file covering compile validation and apply drop/redact behavior.

</details>

<details><summary>Previous review passes</summary>

| commit | score | findings | engine | when |
| --- | --- | --- | --- | --- |
| `1bdacd80303b` | 2/5 | 2 | llm | 2026-09-16 08:52 UTC |

</details>

<sub>Reviewed `1d68a6b12d6d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 30s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw_watcher_window/main.py
```diff
@@ -95,6 +96,9 @@ def main():
             if args.research_enabled
             else None
         )
+        privacy_filter_rules = compile_privacy_rules(
+            getattr(args, "privacy_filter_rules", [])
+        )
         if sys.platform == "darwin" and args.strategy == "swift":
```
TimeToBuildBob, line 102: ❌ **P1** — The privacy filter is applied only in the Python heartbeat_loop path, but on macOS the default strategy is "swift", which spawns a separate Swift binary that sends heartbeats directly to the server. The PR documents this limitation in README, but the config parsing and main() still compile and pass privacy_filter_rules to heartbeat_loop only; the swift path never receives them. A user on macOS who follows the README example and configures privacy_filter rules while using the default strategy will silently have no filtering — sensitive window titles still leave the machine. The README note is a docs workaround, but the feature's core promise ("matching events never leave the machine at all") is broken for the default macOS setup. This is a contract/behavior gap: the config option is accepted but ineffective on a supported platform.

aw_watcher_window/main.py
```diff
@@ -207,7 +217,14 @@ def transform_window(
     exclude_titles=None,
     research_category_map=None,
     research_app_category_map=None,
+    privacy_filter_rules=None,
 ):
+    # Privacy filter runs first — sensitive events never reach other transforms
+    if privacy_filter_rules:
+        current_window = apply_privacy_filters(current_window, privacy_filter_rules)
```
TimeToBuildBob, line 232: ❌ **P1** — In transform_window, the privacy filter runs before the exclude_titles loop, but the exclude_titles loop mutates current_window in place: `current_window["title"] = "excluded"`. When a redact rule matches, apply_privacy_filters returns a shallow copy `result = dict(window)`, so the original current_window is not mutated by the filter itself. However, after the filter returns, the exclude_titles loop operates on the same dict object that was passed in (the original current_window), not on the filtered copy. If a redact rule replaced the title, the exclude_titles loop still sees the original title (because it iterates over the original dict), so a title that was redacted to "excluded" by the privacy filter could still be matched by an exclude_titles pattern and overwritten again — but more importantly, the exclude_titles loop's mutation of the original dict is discarded because transform_window returns the filtered copy, so exclude_titles redaction is silently lost for any event that went through a redact rule. Concretely: with a redact rule matching "bank" and exclude_titles=["secret"], a title "my bank secret" is redacted to "excluded" by the filter, then the exclude_titles loop checks the original title "my bank secret" against "secret", matches, and sets the original dict's title to "excluded" — but the returned dict is the filtered copy whose title is already "excluded", so the result is coincidentally the same. The real break is the reverse: if the redact rule does NOT match but exclude_titles does, the exclude_titles loop mutates the original dict, and that mutation is returned (since apply_privacy_filters returned the original when no rules matched? No — when rules exist but none match, apply_privacy_filters returns `result`, a copy, not the original. So the exclude_titles mutation on the original is lost entirely whenever any privacy rules are configured, even if they don't match. This means exclude_titles stops working whenever privacy_filter_rules is non-empty, a regression for existing users who configure both.

aw_watcher_window/main.py
TimeToBuildBob: Fixed in 1d68a6b. The swift path now fail-closes: if any privacy_filter rule compiled, main() logs an error and sys.exit(1) instead of spawning the binary that would leak titles. Use --strategy jxa or applescript on macOS. Covered by test_swift_refuses_to_start_when_privacy_filter_configured.

aw_watcher_window/main.py
TimeToBuildBob: False positive. transform_window already reassigns `current_window = apply_privacy_filters(...)` before the exclude_titles loop, so the loop mutates the filtered copy and that same object is returned. exclude_titles is not lost when privacy rules are configured.

Regression tests on this head:
- test_exclude_titles_apply_when_privacy_rules_do_not_match
- test_privacy_redact_then_exclude_titles_sees_redacted_copy

aw_watcher_window/main.py
```diff
@@ -95,7 +96,18 @@ def main():
             if args.research_enabled
             else None
         )
+        privacy_filter_rules = compile_privacy_rules(
+            getattr(args, "privacy_filter_rules", [])
+        )
         if sys.platform == "darwin" and args.strategy == "swift":
```
TimeToBuildBob, line 102: ⚠️ **P2** — The swift strategy refusal check only runs when sys.platform == 'darwin' and args.strategy == 'swift', but the default strategy on macOS is 'swift' (config default strategy_macos = "swift"). However, the check is placed after compile_privacy_rules, which is fine. The real issue: the refusal only happens in main() when the Python process starts. But the swift strategy is also the default for macOS, and the README says the watcher refuses to start. The check is correct. Wait, let me look for actual bugs. The privacy filter runs in transform_window, which is only called in the Python heartbeat_loop path. The swift strategy path calls build_swift_command and Popen, and the refusal happens before that. That's fine. But what about the case where privacy_filter_rules is configured but the swift strategy is used on a non-macOS platform? The check is guarded by sys.platform == 'darwin

TimeToBuildBob: False positive — the finding contradicts itself and, by its own text, confirms the code is correct:

> "The check is correct. Wait, let me look for actual bugs. ... the refusal happens before that. That's fine."

The concrete scenario it raises — privacy rules configured with swift strategy on a non-macOS platform — is not a gap: `strategy_macos = "swift"` only applies on macOS; on Linux/Windows the default strategy is the Python path, where the filter runs in `transform_window`. On macOS with `swift`, `main()` refuses to start when rules are compiled (fail-close, added in 1d68a6b), so no unfiltered heartbeat can be sent. No code change needed.

PR #136, opened 2026-08-13
feat(research): add APP_TO_CATEGORY support for non-browser apps

TimeToBuildBob:
## Summary

Implements `APP_TO_CATEGORY` support for Research Edition, closing the spec gap identified in [Matthias's test report (ErikBjare/bob#1108)](https://github.com/ErikBjare/bob/issues/1108).

Non-browser apps previously kept the raw app name while the window title was dropped. Matthias's classifier specifies that app names should be **replaced by broad study categories** (or `Excluded` for unmapped apps), which is strictly more private: unknown apps stop being named at all.

### Python filter (`research_filter.py`)
- New `classify_app(app, app_category_map)` function — case-insensitive exact lookup; returns `"Excluded"` on miss
- `transform()` gains an optional `app_category_map` keyword argument; non-browser branch uses it when supplied, falling back to the previous behaviour (title dropped, raw app name kept) when it is absent

### Config (`config.py`)
- Added `[aw-watcher-window.research_app_category_map]` TOML section to `default_config`
- `parse_args()` loads and exposes it as `parsed_args.research_app_category_map`

### macOS CLI bridge (`macos_cli.py`)
- `build_swift_command()` accepts `research_app_category_map` and passes `--research-app-category app_name category` pairs to the Swift helper

### Swift helper (`macos.swift`)
- Added `researchAppCategoryMap` global and `--research-app-category` argument parsing
- New `classifyApp(_ app: String) -> String` function (case-insensitive lookup, returns `"Excluded"` on miss)
- `applyResearchFilter()` non-browser branch now returns the mapped category when the map is non-empty; falls back to `title: nil` (legacy) when empty

### Heartbeat loop (`main.py`)
- `research_app_category_map` threaded through `heartbeat_loop()`, `build_swift_command()`, and `transform_window()`

### Tests (`tests/test_research_filter.py`)
- `TestClassifyApp` (7 cases): exact match, case insensitivity, exe suffix, spaces, empty map, explicit Excluded entry, unmapped → Excluded
- `TestTransformWithAppMap` (8 cases): mapped, unmapped→Excluded, URL/title stripped, incognito preserved, browser unaffected, legacy (no map) backward compat, explicit Excluded map entry, input not mutated

All 43 tests pass.

## Context

The CI patch script in `ActivityWatch/activitywatch` also needs updating to inject Matthias's 63-entry app map into `[aw-watcher-window.research_app_category_map]` at build time — that change is tracked separately in the activitywatch repo and will land in a follow-up PR once this one merges.

Ref: [ErikBjare/bob#1108](https://github.com/ErikBjare/bob/issues/1108)

### DIFF
--- a/aw_watcher_window/config.py
+++ b/aw_watcher_window/config.py
@@ -11,6 +11,8 @@ strategy_macos = "swift"
 research_enabled = false
 
 [aw-watcher-window.research_category_map]
+
+[aw-watcher-window.research_app_category_map]
 """.strip()
 
 
@@ -73,4 +75,5 @@ def parse_args():
     )
     parsed_args = parser.parse_args()
     parsed_args.research_category_map = dict(config.get("research_category_map", {}))
+    parsed_args.research_app_category_map = dict(config.get("research_app_category_map", {}))
     return parsed_args
--- a/aw_watcher_window/macos.swift
+++ b/aw_watcher_window/macos.swift
@@ -127,6 +127,7 @@ var excludeTitle = false
 var excludeTitlePatterns: [NSRegularExpression] = []
 var researchEnabled = false
 var researchCategoryMap: [(pattern: String, category: String)] = []
+var researchAppCategoryMap: [(app: String, category: String)] = []
 
 let researchBrowserApps = Set([
   "chrome",
@@ -222,6 +223,18 @@ func parseOptionalArguments(_ arguments: ArraySlice<String>) {
       continue
     }
 
+    if argument == "--research-app-category" {
+      let appIndex = arguments.index(after: index)
+      let categoryIndex = appIndex < arguments.endIndex ? arguments.index(after: appIndex) : arguments.endIndex
+      guard appIndex < arguments.endIndex, categoryIndex < arguments.endIndex else {
+        error("Missing app/category values for --research-app-category")
+        exit(1)
+      }
+      researchAppCategoryMap.append((app: arguments[appIndex], category: arguments[categoryIndex]))
+      index = arguments.index(after: categoryIndex)
+      continue
+    }
+
     error("Unknown argument: \(argument)")
     exit(1)
   }
@@ -255,6 +268,18 @@ func classifyResearch(_ title: String, url: String?) -> String {
   return "excluded"
 }
 
+func classifyApp(_ app: String) -> String {
+  // Case-insensitive exact lookup of app name in the app category map.
+  // Returns the mapped category, or "Excluded" when the app is not in the map.
+  let appLower = app.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+  for item in researchAppCategoryMap {
+    if item.app.lowercased() == appLower {
+      return item.category
+    }
+  }
+  return "Excluded"
+}
+
 func applyResearchFilter(_ data: NetworkMessage) -> NetworkMessage {
   if !researchEnabled {
     return data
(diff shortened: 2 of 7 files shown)


ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Fixed the three Greptile-flagged defects in `121cea8`:

1. **Privacy** — non-browser app names were retained in `app` while the category was written to `title`, leaking raw app identities. The category now **replaces** `app` and the title is dropped, matching the documented anonymization contract ("otherwise replaced with 'Excluded'").
2. **Empty map treated as active** — an empty `research_app_category_map` (the default when `research_enabled` but no map configured) classified every non-browser app as "Excluded". Empty map now falls back to legacy behaviour (keep app name, drop title), consistent with Swift.
3. **Case-insensitive contract** — `classify_app` assumed lowercased map keys, so mixed-case configured keys missed. Lookup now normalizes both sides.

Swift `applyResearchFilter` updated to match. Tests updated to codify the corrected behaviour (including new cases for empty-map fallback and mixed-case keys); `pytest tests/` → 59 passed.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the Swift/Python normalization parity gap in `9e381ae`.

`classify_app` (Python) trimmed **both** the incoming app name and the configured map key (`map_app.strip().lower()`), while Swift `classifyApp` trimmed only the incoming name (`item.app.lowercased()`). A configured entry carrying stray whitespace — plausible given the map is injected into TOML at build time from a 63-entry dict — would classify correctly on Linux/Windows and silently fall through to `"Excluded"` on the macOS Swift path for the same config.

This is load-bearing for the study: the Research Edition participant who reported the original issue is on macOS, so the Swift path is the hot path there. Both sides now trim and lowercase.

aw_watcher_window/config.py
```diff
@@ -11,6 +11,8 @@
 research_enabled = false
 
 [aw-watcher-window.research_category_map]
+
+[aw-watcher-window.research_app_category_map]
```
ErikBjare, line 15: Why add both these options/configs? Shouldn't they be a single option? And we already have `research_enabled = false`? Why are they in default config at all? (dev/research options shouldn't get persisted as default/initial config)

TimeToBuildBob: Fixed in #137.

**Why they were in default config at all** — that was the real bug, and worse than it looked: `load_config_toml()` comments out *keys* but not *table headers*, so the two `[aw-watcher-window.research_*_category_map]` tables landed **uncommented** in every fresh install's config file. They were the only live lines in it. Now moved to a separate template that is read but never written to disk; a fresh config has zero research keys (verified by running against an empty config dir, not by reading the diff).

**Should they be one option** — I think not, because the two maps match differently: `research_category_map` is a case-insensitive *substring* match on browser URL/title, `research_app_category_map` is an *exact* match on app name. One table would make a key like `"mail"` both at once. Also `research_enabled` is the switch and the maps are its data, so not redundant. Say the word if you still want them collapsed — I'd pick one matching rule for both first.

One thing worth knowing: the Research Edition build seds this exact file from the superproject (`s/^research_enabled = false$/research_enabled = true/`). Kept the target byte-identical and pinned it with a test, since that coupling is cross-repo and would break silently.

Issue #139, opened 2026-08-26
[macOS] aw-watcher-window-macos leaks AXObserver/Mach ports, causing high wired memory


MR-HOWE:
## Summary

On macOS, `aw-watcher-window-macos` appears to leak Accessibility (`AXObserver`) receive ports during long-running use.

After ActivityWatch 0.13.2 had been running for about 16 days, the helper owned **7,339 Mach receive ports** containing **3,791,856 queued messages**. This represented **99.897% of all queued Mach messages on the system**.

Quitting ActivityWatch caused approximately **3.807 million related IPC, VM map copy, VM object, and compressor objects to disappear immediately**, confirming that the accumulated kernel objects belonged to the ActivityWatch window watcher.

The accumulation resulted in approximately **6.1 GiB of wired memory** on a Mac with 18 GiB unified memory.

## Environment

- ActivityWatch: 0.13.2
- macOS: 26.5.2 (25F84)
- Hardware: Apple M3 Pro, 18 GiB unified memory
- Window watcher strategy: Swift
- Helper: `aw-watcher-window-macos`
- Helper architecture: x86_64 running through Rosetta
- Uptime when diagnosed: approximately 16 days

I have not yet tested the current prerelease build.

## Observed behavior

Wired memory remained around 6.1–6.3 GiB even when no VM, container, local LLM, or other obvious high-wired workload was running.

`zprint` showed several related kernel zones with approximately 3.8–4.8 million live objects:

```text
VM.map.entries           4,840,900
VM.map.copies            3,832,294
ipc.kmsgs                3,815,571
vm.objects               4,271,204
data.kalloc.192          3,767,849
compressor_pager         4,068,152
compressor_slots.64      3,877,006
```

Between two samples taken approximately five hours apart, the following zones grew by nearly identical amounts:

```text
ipc.kmsgs          +50,788
VM.map.copies      +49,823
data.kalloc.192    +50,661
```

A 30-second idle sample was flat, suggesting that the accumulation is event-triggered, possibly by switching applications or focused windows, rather than a fixed-rate background loop.

## Mach port evidence

I captured a system-wide Mach port snapshot using Apple's `lsmp`:

```bash
sudo /usr/bin/lsmp -a -j /tmp/lsmp.json
```

The system contained 3,795,755 queued messages in total. The top process was:

```text
Process:                   aw-watcher-window-macos
PID:                       4669
Total port names:          7,428
Receive ports:             7,339
[...]

PR #141, opened 2026-08-30
macOS: keep foreground tracking coherent when browser context fails

hawai-i:
## Summary

Browser URL/title enrichment is optional context, but the macOS Swift watcher currently returns without emitting a foreground heartbeat when some Chrome/Safari ScriptingBridge lookups fail. Heartbeat merging then extends the previous application's event across the real browser interval.

This change:

- emits foreground app/title evidence even when browser context acquisition fails;
- centralizes activation and polling through one PID reconciliation path;
- binds the tracked application, AX observer, focused window, and emitted app identity to the same PID;
- repairs a missing/stale observer and rejects stale AX callbacks;
- consumes the activated `NSRunningApplication` from the workspace notification;
- safely removes the previous title notification and run-loop source.

## Validation

A controlled `Codex → Chrome → Finder → Chrome → Codex` sequence was scored against WindowServer's top visible layer-0 owner. Browser context failure was forced at the existing pre-heartbeat failure point in both the before and fixed test builds.

| Metric | Before | After |
| --- | ---: | ---: |
| Overall foreground accuracy | 41.9460% | 99.8482% |
| Chrome accuracy | 0.0000% | 99.8799% |
| Boundaries within ±3 s | 2/4 | 4/4 |
| Unmatched time | 60.811 s | 0.159 s |

A normal companion build also retained same-title Chrome `page-a → page-b` URL changes.

Automated Swift state tests cover:

- PID change → observer rebuild;
- same PID → focused-window refresh;
- missing observer → repair;
- stale callback → reject;
- callback for the current PID → accept.

`make build-swift`, `make test-swift`, a real PyInstaller 6.11 package build, `mypy`, and `git diff --check` passed. The packaged arm64 helper preserves the macOS 12.0 minimum. The borrowed local Python environment reported 61 passing tests and two macOS config-path failures that reproduce unchanged on the base commit.

## Related work

[ActivityWatch/activitywatch#1411](https://github.com/ActivityWatch/activitywatch/pull/1411) grants the responsible outer app Apple Events permission, which lets browser context acquisition succeed. This PR is logically independent: foreground truth must still be emitted when optional context is unavailable.

This also corrects the mismatched title-notification cleanup and observer teardown discussed in #139, but this PR does not claim a separate multi-day leak validation.


ErikBjare:
@TimeToBuildBob build on top of this to resolve conflict and fix Greptile concerns. Make sure to keep attribution.

TimeToBuildBob:
Rebased onto current master (after #145) and opened #146.

Greptile P1: Chrome ScriptingBridge failure now keeps app identity but drops title and URL, since that lookup is the only incognito detector.

Greptile P2: `make test-swift` compiles with the macOS 12 deployment target.

Original work stays attributed to hawai-i (`Co-Authored-By` on the rebase commit). Leaving this PR open until #146 merges.

PR #142, opened 2026-09-06
fix(research): recognize original browser aliases

TimeToBuildBob:
## Summary

- restore the Arc, Vivaldi, and Windows Chromium aliases from Matthias's original study classifier
- keep the Python and Swift Research Edition browser sets identical
- verify those browsers retain their app identity, store the study category, and drop the URL

Companion to ActivityWatch/activitywatch#1431. That retained-app change leaves the watcher app-category map empty; this PR ensures these proven browser identities still take the browser categorization path.

## Verification

- `poetry run python -m pytest tests/test_research_filter.py tests/test_main.py -q` — 64 passed
- `make test` — 70 passed; mypy clean


TimeToBuildBob:
@greptileai review

Issue #101, opened 2024-02-14
[macos] Process crashes without error


ErikBjare:
Watcher process suddenly died after running okay for weeks. Lost 2h of data.

Last messages:

```
2024-02-14 11:50:18 [aw-watcher-window-macos] [DEBUG] [heartbeat] bucket: aw-watcher-window_erb-m2.localdomain, timestamp: 2024-02-14 10:50:18 +0000, pulsetime: 1.7, app: Alacritty, title: [0:mono-gpt-engineer] ~/P/m/app (/Users/erb/Programming/mono-gpt-engineer/app), url:
2024-02-14 11:50:18 [aw-watcher-window-macos] [DEBUG] sending old heartbeat for merging
2024-02-14 11:50:23 [DEBUG]: queue path '/Users/erb/Library/Application Support/activitywatch/aw-client/queued/aw-watcher-window.v1.persistqueue'  (aw_client.client:418)
⏎
```

Exit code 0.

BelKed:
Started to happen to me also... Not sure what's happening here :(
Perhaps it could be fixed by #105?

claudiu-bargianu:
The thing crashed for me mid work day too. Happened yesterday on 23feb and today. Tahoe 26.3

TimeToBuildBob:
This is a tricky one — exit code 0 with no error output suggests the process received a clean signal (SIGTERM) or hit an unhandled nil/optional that caused an immediate exit without a crash report.

Given the symptoms (silent death after running for hours/days, last output is a normal heartbeat), a few possible causes:

1. **macOS process management**: macOS can terminate processes it deems inactive or resource-heavy, especially during thermal throttling or memory pressure. This usually happens silently with exit code 0. Check Console.app → system.log around the time of the crash for any `jetsam` or `thermalmonitor` entries.

2. **Swift nil unwrap in the watcher binary**: The related issue #100 reports explicit nil unwrap crashes. If the watcher encounters a window state that triggers a nil unwrap in a non-debug build, it could exit cleanly depending on how the error propagates. PR #105 (improved handling of missing application attributes) may address this.

3. **aw-client queue path message**: The last log line shows the client accessing its persistent queue, which happens when the server connection drops. If the client's reconnection logic fails fatally, it could explain the exit.

[...]

PR #145, opened 2026-09-15
fix(macos): survive non-string AX titles and stop leaking AXObserver

### LINKED ISSUE #139: [macOS] aw-watcher-window-macos leaks AXObserver/Mach ports, causing high wired memory

### LINKED ISSUE #144: macOS: native helper crashes when NSHTTPURLResponse is bridged as a window-title string
### Environment

- ActivityWatch: `0.14.0b5`
- Installation: Homebrew `activitywatch@beta`
- macOS: `15.7.3 (24G419)`
- Architecture: Apple Silicon, native arm64 (`translated: false`)
- Hardware model: `Mac15,6`

### Problem

After running normally for approximately four days, `aw-watcher-window-macos` crashed while handling a focused application/window change. Window tracking stopped, while ActivityWatch, `aw-watcher-afk`, and `aw-sync` continued running.

The last window event was recorded at `2026-09-14 17:33:53 +0530`. The crash occurred at `17:33:55`.

### Crash

The macOS crash report shows:

```text
NSInvalidArgumentException:
-[NSHTTPURLResponse length]: unrecognized selector sent to instance
```

Relevant stack frames:

```text
static String._unconditionallyBridgeFromObjectiveC(_:)
MainThing.windowTitleChanged(_:axElement:notification:)
MainThing.focusedWindowChanged(_:window:)
MainThing.focusedAppChanged()
```

The process terminated with `EXC_CRASH / SIGABRT` on the main thread. This appears to mean that an Objective-C API returned an `NSHTTPURLResponse` where the Swift code expected a string, and the value was bridged without validating its runtime type.

### Restart behavior

The Tauri manager logged:

```text
[2026-09-14 17:33:55][INFO][aw_tauri_lib::manager] Stopped module: aw-watcher-window
[2026-09-14 17:33:55][INFO][aw_tauri_lib::manager] Module aw-watcher-window exited successfully
```

The native helper crashed, but the Python wrapper apparently exited with status 0. The manager therefore considered it a successful exit and did not restart the window watcher.

The silent exit/restart symptom resembles #101. The one-shot helper lifecycle and failure to restart are also discussed in #139, but the `NSHTTPURLResponse` exception appears to be a separate crash cause.

### Expected behavior

1. Unexpected Accessibility or ScriptingBridge return types should be handled without crashing.
2. If the native helper crashes, the wrapper should return a failure status so the module manager can restart it.

### Reproduction

No reliable reproduction yet. It occurred during an application/window focus change after several days of normal operation.

I have a macOS `.ips` crash report, but have not attached the raw file because it contains persistent device/session identifiers. I can provide a sanitized version if useful.

TimeToBuildBob:
## Summary

macOS window tracking dies in two related ways after a few days:

1. **Crash (#144).** `windowTitleChanged` bridged `kAXTitleAttribute` with `as? String`. That path calls `-[obj length]` without a class check, so a non-NSString (here `NSHTTPURLResponse`) aborts the helper with `NSInvalidArgumentException`.
2. **Leak (#139).** `focusedAppChanged` removed the run-loop source and then let `AXObserverCreate` overwrite the previous observer pointer without `CFRelease`. The old Mach receive port kept filling to `qlimit` (1024). Independently, `focusedWindowChanged` unregistered `kAXFocusedWindowChangedNotification` from the old window even though it had registered `kAXTitleChangedNotification`.
3. **Silent stop.** The Python wrapper ignored the helper's wait status, so a SIGABRT still looked like a clean exit to aw-tauri and window tracking stayed off.

## Changes

- Read AX strings/elements via CF type ID (`axString` / `axElement`) instead of Swift's NSString bridge.
- Tear down the previous observer: unregister title + focused-window notifications, drop the run-loop source, nil the refs, then create a new observer into a fresh out-pointer.
- Map helper signal deaths (`-6` → 134) to a non-zero wrapper exit so the module manager restarts the watcher.

Observer teardown follows the patch Gassandrid measured on #139 (200/2000 app-switch iterations, no port growth). I cannot compile the Swift helper on this Linux host; Python tests cover the exit-status path.

## Related

- Fixes #144
- Fixes #139
- Related: #101 (silent successful-looking exit after a helper crash)
- Related: #141 (larger foreground-reconciliation rewrite; still worth landing separately — it does not type-check AX strings, and its teardown does not unregister the app-level focused-window notification)

## Test plan

- [x] `poetry run python -m pytest tests/test_main.py` (19 passed)
- [x] `poetry run mypy aw_watcher_window/ --ignore-missing-imports`
- [ ] On macOS: `make build-swift`
- [ ] Switch apps rapidly and confirm `aw-watcher-window-macos` Mach port count stays bounded (`sudo lsmp` / `sudo zprint`)
- [ ] Confirm a helper abort now makes `aw-watcher-window` exit non-zero so aw-tauri restarts it


ErikBjare:
@greptileai review

ErikBjare:
@TimeToBuildBob review

TimeToBuildBob (commented):
Reviewed against ActivityWatch/aw-watcher-window#144, #139 (Gassandrid's measured teardown), and aw-tauri's module manager. I authored this PR; this is the adversarial pass.

**Verdict:** land it. No blocking issues. GitHub will not let me approve my own PR.

### Crash path (#144)

`as? String` on an AX value really does go through `String._unconditionallyBridgeFromObjectiveC` → `-[obj length]` with no class check. `NSHTTPURLResponse` is exactly that trap. `axString` / `axElement` via `CFGetTypeID` is the right fix, and every AX title/role/focused-window read on this path is converted.

The leftover `urlRef as? NSURL` is a different bridge (class-checked) and is the URL success path; `axString` is only the CFString fallback.

### Leak path (#139)

`tearDownObserver` matches the patch Gassandrid measured at 200 and 2,000 app-switch iterations (no Mach-port growth):

1. Unregister `kAXTitleChangedNotification` on `oldWindow`
2. Unregister `kAXFocusedWindowChangedNotification` on `observedApp`
3. Drop the run-loop source
4. Nil the refs so Swift ARC releases the `AXObserver` (the Mach-port owner)
5. Create into a fresh out-pointer, not `&observer`

Two extras vs that patch, both correct: `deinit` also tears down, and `pollActiveWindow` no longer force-unwraps a nil observer. `RunLoop.current` for both add and remove is consistent — mixing `main`/`current` (as the #139 sketch did) would leak if those ever diverged.

### Exit status

`swift_helper_exit_status(-6) == 134` is the mapping `Popen.wait()` uses for SIGABRT. aw-tauri restarts only when `!output.status.success()` (`manager.rs`); exit 0 was the silent-off. `stop_module` sets `modules_pending_shutdown` before SIGTERM, so mapping helper SIGTERM → 143 will not restart during quit. aw-qt already treats any death-while-started as unexpected (dialog, no auto-restart) — this change is for the tauri path.

Tests cover the mapping and the swift-strategy `SystemExit(134)` path.

### CI

macos-latest already compiled this:

```text
swiftc -target arm64-apple-macosx12.0 aw_watcher_window/macos.swift
```

Succeeded. The two warnings (optional interpolation in the Chrome/Safari title-diff logs) are pre-existing. The unchecked `make build-swift` box in the PR body is stale.

### Out of scope

- #141 also rewrites observer lifecycle and will conflict. Land this first: it actually type-checks AX strings and unregisters the app-level focused-window notification; #141 does neither, and still has `windowTitle as? String`.
- Chrome/Safari ScriptingBridge failure still `return`s without a heartbeat. That's #141, not this PR.
- Remaining live check I cannot do here: Mach-port count under rapid app switching (`sudo lsmp`). Gassandrid already ran that harness against this teardown; CI compiled the helper but did not rerun it.

No code changes requested.

PR #146, opened 2026-09-15
fix(macos): keep foreground tracking when browser context fails

TimeToBuildBob:
## Summary

Rebase of #141 (hawai-i) onto current `master` after #145. Original work is preserved and attributed (`Co-Authored-By: hawai-i`).

Browser URL/title enrichment is optional, but the macOS Swift watcher used to return without a heartbeat when Chrome/Safari ScriptingBridge failed. Heartbeat merging then extended the previous application across the real browser interval.

This keeps #141's PID-based foreground reconciliation:

- emit a foreground heartbeat even when browser context acquisition fails
- bind tracked app, AX observer, focused window, and emitted app identity to one PID
- repair a missing/stale observer and reject stale AX callbacks
- consume the activated `NSRunningApplication` from the workspace notification

## Conflict with #145

#145 landed on `master` after #141 opened. The rebase keeps that work:

- `axString` / `axElement` CF type checks (the #144 `NSHTTPURLResponse` crash)
- observer teardown that unregisters both title and focused-window notifications and tracks `observedApp` (the #139 Mach-port leak)

Unsafe `as? String` / `as! AXUIElement` casts from #141 are not carried forward.

## Greptile on #141

**P1 (incognito):** ScriptingBridge is the only Chrome incognito detector. The original fallback sent AX-derived app + title when that lookup failed. The rebase still emits the Chrome app identity (so tracking stays coherent) but drops title and URL. Successful incognito detection still blanks both app and title.

**P2 (tests):** `make test-swift` now compiles with `-target …-macosx$(MACOSX_DEPLOYMENT_TARGET)` (12.0), matching the production helper.

## Validation

Python tests: 77 passed. mypy clean. Swift helper/tests need Darwin CI (`macos-latest`).

## Related

- #141 — original PR (leave open until this lands)
- #145 / #144 / #139 — AX title crash and observer leak, now on master


TimeToBuildBob:
macos-latest failed at `make build-swift`: compiling `macos.swift` together with `macos_state.swift` rejects top-level expressions (`formatter.formatOptions = …`, `start()`, `RunLoop.main.run()`).

The rebase onto master after ActivityWatch/aw-watcher-window#145 dropped the ActivityWatch/aw-watcher-window#141 `@main` wrapper and closure inits. Restored in 8ac81d3. CI is re-running.

TimeToBuildBob:
CI-green and mergeable — **waiting only on a maintainer click**.

This PR is ready to merge, but the bot has pull-only access to this repo and can't self-merge — surfacing it here so it isn't lost. The monitoring loop will stop re-flagging it now that this note is posted.

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 2 findings disposed (rejected; fixed in 90aad7fbb413)

<sub>Updated after inline dispositions on finding threads — this is the current state; the verdict below is frozen at review time and is kept as the historical record of that pass.</sub>

<details><summary>Finding disposition</summary>

| Finding | Severity | State |
| --- | --- | --- |
| `aw_watcher_window/macos.swift:211` | P0 | rejected — The verification step claimed "lines 192-209 contain top-level let main/formatter/encoder" and inferred "'main' attribut |
| `aw_watcher_window/macos.swift:682` | P1 | fixed in `90aad7fbb413` |
| `aw_watcher_window/macos.swift:756` | P2 | superseded by latest review (not reproduced) |
| `aw_watcher_window/macos.swift:819` | P2 | superseded by latest review (not reproduced) |
| `aw_watcher_window/macos.swift:746` | P2 | accepted-tradeoff |

</details>




### Needs a look — P2 only

Confidence 4/5

<sub>ℹ️ Consensus was degraded on this run: 2 of 3 passes answered, so findings were filtered at 2-of-2 agreement rather than 2-of-3 — less filtered than usual; 2 of 3 fan-out jobs answered, so the findings above were filtered against less evidence than the threshold assumes.</sub>

<details><summary><b>1 finding</b> · ⚠️ 1 P2</summary>

**⚠️ P2 medium** — `aw_watcher_window/macos.swift:746`

In `updateFocusedWindow`, when a new focused window is registered, the code attempts `AXObserverAddNotification` for the window's title. If that call fails transiently (e.g., the window is not yet ready), it logs the error but continues: `oldWindow` is set to the new window and a heartbeat is emitted. On subsequent polls, `refreshFocusedWindow` calls `updateFocusedWindow` again; because `oldWindow` now equals the current window, `windowChanged` becomes false and the add is not re-attempted. Thus the title-change observer is never installed for that window, so live title/URL updates stop being received until the focused window changes or the AX observer is rebuilt. The watcher still sends heartbeats on each poll (every 10s) with the current title read via AX, so data is not lost but is delayed by up to 10 seconds, and tab switches that occur between polls are missed until the next poll.

```suggestion
// Release the observer we created but could not install.
      // AXObserverCreate returns a +1 retained object; use CFRelease to avoid leaking.
      CFRelease(newObserver)
      return
```

<sub>**How this was verified:** Checked the AXObserverCreate signature in the macOS SDK: it is a C function returning AXError, with the observer as an out-parameter, not marked CF_RETURNS_RETAINED. The code's own comment acknowledges the ARC ambiguity. The leak path is reached whenever addResult is neither success nor notificationAlreadyRegistered.</sub>

<sub>Consensus: 2/2 passes agreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

<details><summary>Previous review passes</summary>

| commit | score | findings | engine | when |
| --- | --- | --- | --- | --- |
| `8ac81d39c2f8` | 4/5 | 2 | llm | 2026-09-16 08:53 UTC |
| `9424b4f6a054` | 1/5 | 3 | llm | 2026-09-16 12:44 UTC |

</details>

<sub>Reviewed `90aad7fbb413` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 594s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw_watcher_window/macos.swift
```diff
@@ -721,84 +637,222 @@ class MainThing {
...
+
+  func handleAXNotification(
+    observer callbackObserver: AXObserver,
+    element: AXUIElement,
+    notification: CFString
+  ) {
+    guard let application = foregroundApplication,
+          let currentObserver = observer,
+          CFEqual(callbackObserver, currentObserver),
+          axCallbackBelongsToForeground(trackedPID: trackedPID, elementPID: elementPID(element)) else {
```
TimeToBuildBob, line 764: ⚠️ **P2** — handleAXNotification accepts any title-change callback whose observer matches the current one and whose element's PID equals the tracked PID. When a user switches between two windows of the same application, updateFocusedWindow removes the old window's title notification and registers the new one, but a title-change event already queued for the old window can be processed after this switch. The guard passes because both windows belong to the same PID and the observer is unchanged, so the callback reaches the kAXTitleChangedNotification branch and calls emitHeartbeat with the stale window. The watcher then records the previous window's title/URL as if it were the current foreground window, producing an incorrect heartbeat.

aw_watcher_window/macos.swift
```diff
@@ -721,84 +637,222 @@ class MainThing {
...
+         let windows = safariObject.windows,
+         let frontWindow = windows().first,
+         let activeTab = frontWindow.currentTab {
+        // Safari doesn't allow incognito mode to be inspected, so we do not know if we should hide the url
+        data.url = activeTab.URL
+        if let tabTitle = activeTab.name, tabTitle != "", data.title != tabTitle {
+          data.title = tabTitle
+        }
+      } else {
+        log("Failed to read Safari context; emitting foreground heartbeat without URL")
```
TimeToBuildBob: ⚠️ **P2** — When Safari ScriptingBridge lookup fails, the else branch at lines 818-820 only logs and leaves `data` as built from the AX window title (and no URL). Unlike the Chrome failure branch, which replaces `data` with `chromeHeartbeatAfterContextFailure` dropping title and URL, the Safari path emits a heartbeat carrying the raw AX title. In Safari private browsing, the AX window title can be the page title (URLs are hidden from accessibility, but titles are not), so this transmits potentially sensitive page titles to the server. The previous code returned without a heartbeat when ScriptingBridge failed, so this behaviour is new and introduces a privacy leak.

aw_watcher_window/macos.swift
TimeToBuildBob: Fixed in 9424b4f6a054 — title-change callbacks now also require the element to be the focused window (`axTitleCallbackBelongsToFocusedWindow` + `CFEqual` against `oldWindow`). A queued title event from a previous window of the same PID is dropped.

aw_watcher_window/macos.swift
TimeToBuildBob: Fixed in 9424b4f6a054 — Safari ScriptingBridge failure now uses the same keep-app / drop-title-URL fallback as Chrome (`browserHeartbeatAfterContextFailure`). Foreground tracking stays coherent without emitting the AX page title.

aw_watcher_window/macos.swift
```diff
@@ -192,18 +192,29 @@ let researchBrowserApps = Set([
...
+  let encoder = JSONEncoder()
+  encoder.dateEncodingStrategy = .custom({ date, encoder in
+    var container = encoder.singleValueContainer()
+    let dateString = formatter.string(from: date)
+    try container.encode(dateString)
+  })
+  return encoder
+}()
+
+@main
```
TimeToBuildBob, line 214: 🛑 **P0** — The new test target compiles macos_state.swift and tests/macos_state_tests.swift together, but macos_state.swift imports Foundation and defines functions that are also used by macos.swift. The test binary does not include macos.swift, so it only tests the pure helpers. That is fine. However, the Makefile's test-swift target uses `swiftc -target ...` without `-parse-as-library`, and the test file uses @main. In Swift, a file with top-level code (macos.swift has top-level `let main = MainThing()` and `let formatter` etc.) cannot be compiled with @main in the same module, but the test target only compiles macos_state.swift and the test file, so no conflict. The production build compiles macos.swift and macos_state.swift together; macos.swift has top-level code and also @main. In Swift, a file with top-level code cannot also have @main? Actually, a file with top-level code is treated as the main entry point, and @main is not allowed in the same module. The original macos.swift had top-level code (let main = MainThing(), start(), RunLoop.main.run()). The PR changed it to @main struct and removed the top-level start()/RunLoop calls, but still has top-level `let main = MainThing()` and `let formatter` etc. In Swift, top-level code in a file is only allowed in a file named main.swift, or when compiling with -parse-as-library? Actually, top-level code is allowed in any file when the file is the main file, but if you have @main, you cannot have top-level executable code. The Swift compiler will error: 'main' attribute cannot be used in a module that contains top-level code. This is a compile error. The PR's build-swift target compiles macos.swift and macos_state.swift together; macos.swift has top-level `let main = MainThing()` and `let formatter` etc. That will fail to compile with @main. I need to verify: In Swift, top-level code is allowed only in a file named main.swift, or when the file is passed as the first file? Actually, the rule is: top-level code is allowed in a file named main.swift, or in a file that is the only file? Let me recall: Swift allows top-level statements in a file named main.swift, and in files compiled as the main file. If you use @main, you cannot have top-level statements. The error is: 'main' attribute cannot be used in a module that contains top-level code. So this PR will not compile. The original code had top-level code and no @main. The PR added @main but left top-level `let main = MainThing()` and `let formatter` etc. That is a compile error. This is a P0 because the build breaks. I should report this.

aw_watcher_window/macos.swift
```diff
@@ -721,84 +637,238 @@ class MainThing {
...
     let selfPtr = UnsafeMutableRawPointer(Unmanaged.passUnretained(self).toOpaque())
-    AXObserverAddNotification(
-      newObserver, focusedApp, kAXFocusedWindowChangedNotification as CFString, selfPtr)
+    let addResult = AXObserverAddNotification(
+      newObserver,
+      focusedApp,
+      kAXFocusedWindowChangedNotification as CFString,
+      selfPtr
+    )
+    guard addResult == .success || addResult == .notificationAlreadyRegistered else {
```
TimeToBuildBob, line 684: ❌ **P1** — In rebuildObserver, when AXObserverAddNotification fails (other than .notificationAlreadyRegistered), the function logs and returns without setting observer/observedApp, but foregroundApplication was already set to the new application and the newly created AXObserver is leaked (never added to the run loop, never released). More importantly, the caller reconcileForegroundApp then proceeds to call refreshFocusedWindow, which uses the stale observer (nil) and the new foregroundApplication. Since observer is nil, updateFocusedWindow's windowChanged branch is skipped, but oldWindow is still set to the focused window and emitHeartbeat is called. The next poll sees observer == nil and rebuilds, but the immediate heartbeat is emitted with the new app identity while oldWindow may still reference the previous app's window. The observable consequence is a heartbeat with the new app but a stale window title/URL from the previous app, and a leaked AXObserver each time the focused-window notification registration fails. The failure path is not rare: apps that do not expose a focused-window AX attribute (e.g. some menu-bar-only apps) cause AXObserverAddNotification to return an error, so this can happen on every activation of such an app.

aw_watcher_window/macos.swift
TimeToBuildBob: False positive — this compiles. File-level `let main` / `formatter` / `encoder` are declarations, not executable top-level statements. `@main` owns `start()` / `RunLoop.main.run()` because compiling `macos.swift` together with `macos_state.swift` is not a script compilation unit (restored in 8ac81d3 after macos-latest rejected the top-level assignments).

macos-latest already ran `swiftc -target arm64-apple-macosx12.0 aw_watcher_window/macos.swift aw_watcher_window/macos_state.swift` on this `@main` shape and passed. Removing `@main` would re-break that two-file build.

aw_watcher_window/macos.swift
TimeToBuildBob: Fixed in 90aad7fbb413 — `rebuildObserver` no longer binds `foregroundApplication` before observer install. On `AXObserverAddNotification` failure the unused observer is parked on the managed `observer` property and niled so Swift ARC releases it (same contract as the Create-overwrite warning; `CFRelease` would double-free). The caller still sets identity so a heartbeat is emitted; the next poll retries because `observer` stays nil.

aw_watcher_window/macos.swift
```diff
@@ -721,84 +640,245 @@ class MainThing {
...
+      }
+      if let window = window {
+        let selfPtr = UnsafeMutableRawPointer(Unmanaged.passUnretained(self).toOpaque())
+        let addResult = AXObserverAddNotification(
+          observer,
+          window,
+          kAXTitleChangedNotification as CFString,
+          selfPtr
+        )
+        if addResult != .success && addResult != .notificationAlreadyRegistered {
```
TimeToBuildBob, line 746: ⚠️ **P2** — In `updateFocusedWindow`, when a new focused window is registered, the code attempts `AXObserverAddNotification` for the window's title. If that call fails transiently (e.g., the window is not yet ready), it logs the error but continues: `oldWindow` is set to the new window and a heartbeat is emitted. On subsequent polls, `refreshFocusedWindow` calls `updateFocusedWindow` again; because `oldWindow` now equals the current window, `windowChanged` becomes false and the add is not re-attempted. Thus the title-change observer is never installed for that window, so live title/URL updates stop being received until the focused window changes or the AX observer is rebuilt. The watcher still sends heartbeats on each poll (every 10s) with the current title read via AX, so data is not lost but is delayed by up to 10 seconds, and tab switches that occur between polls are missed until the next poll. _(2/2 passes)_

TimeToBuildBob: Accepted trade-off, not a regression of this PR's goal. If title-notification add fails once, `windowChanged` becomes false and we do not retry the add until the focused window changes. Poll still emits a heartbeat every 10s with the current AX title, so the title is not lost — only live title-change callbacks between polls. Retrying a failed title observer is a follow-up, not required to keep foreground tracking when browser ScriptingBridge fails.

TimeToBuildBob:
@greptileai review

aw_watcher_window/macos.swift
```diff
@@ -721,84 +640,245 @@ class MainThing {
...
+          log("Failed to observe title changes for pid \(application.processIdentifier): \(addResult.rawValue)")
+        }
+      }
+    }
+
+    oldWindow = window
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in fe2f5c1c7651 — `updateFocusedWindow` now tracks `titleNotificationRegistered`. If `AXObserverAddNotification` for `kAXTitleChangedNotification` fails, the focused window is still stored (poll heartbeats keep the current AX title) but registration stays retryable on the next poll. `shouldAttemptTitleNotificationRegistration` is the decision; Swift state tests cover the retry case.

TimeToBuildBob:
@greptileai review

